PR #1, opened 2023-02-24
Init network package, basic query

tanlethanh:
I have tested the network package on independent repo, and it's work fine.
Some troubles:
- I cannot rebuild background worker script
- Websocket didn't work in background worker


fluctlight-kayaba (approved):
next time, avoid push comments - this is initial commit while we're scaffolding stuff then its fine

PR #3, opened 2023-02-27
create alias hook, add Passcode screen, minor UI update, full (dummy) flow from login to dashboard

FME849:


apps/extension/src/screens/Login/index.tsx
```diff
@@ -4,6 +4,7 @@ import {
 	useSharedValue,
 	withTiming,
 } from 'react-native-reanimated';
+import { useNavigate } from 'react-router-dom';
```
fluctlight-kayaba, line 7: should alias to `utils/hook`

fluctlight-kayaba (approved):

PR #28, opened 2023-03-07
create component request and page connection request

SuperDMWebDev:


fluctlight-kayaba:
@SuperDMWebDev you need to enable `eslint` and fix all warnings first.

apps/extension/src/components/Request/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  buttonPrimaryContent: string;
+}
+interface RuleProps {
+  icon: string;
+  description: string;
+}
+const Request: FC<RequestProps> = ({ type, rules, buttonPrimaryContent }) => {
+  return (
+    <View className="request__container w-screen h-screen">
+      <View className="request__header h-[42px] relative">
```
fluctlight-kayaba, line 18: seems that you're using `css` for styling? we don't use direct `css`, just `tailwind-css` classNames or [Stylesheet.create](https://reactnative.dev/docs/stylesheet#create).

apps/extension/src/components/Request/index.tsx
```diff
@@ -0,0 +1,160 @@
+import { View, Image, Text, TouchableOpacity } from '@walless/ui';
+import { FC } from 'react';
+import styles from './style';
```
fluctlight-kayaba, line 3: There is no `./style` file, and we don't use/import `css` at component level, in fact don't use `css` directly.

apps/extension/src/components/Request/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+        {type == requestType.connectionRequest ? (
+          <View className="request__rule mx-[23px] bg-[#01131F] rounded-[8px] border border-solid border-[#01131F] px-[8px] py-[5px] mb-[40px]">
+            <Text
+              style={{ color: '#587A90' }}
+              className="px-[20px] text-center pb-[10px] request__rule-header weight-[400] text-[12px]"
+            >
+              Under Realm would like to connect with your Walless account to:
+            </Text>
+            <View className="mb-[10px] bg-[#2A4C63] rotate-[0.07deg] w-full h-[1px] mx-auto "></View>
```
fluctlight-kayaba, line 63: too much `[..]` style, please use closest named sizes like `mb-3`(which is `12px`) instead.

apps/extension/src/components/Request/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+          }}
+        >
+          <Text
+            style={{ textAlign: 'center', lineHeight: 15, color: '#88c3c0' }}
+          >
+            Deny
+          </Text>
+        </TouchableOpacity>
+      </View>
+    </View>
```
fluctlight-kayaba, line 156: Too complex, conditional check. Split it into smaller child component instead. Keep a component dead simple with less than `120` lines if possible. Debug/refactor a huge component took us more time to keep it small.

PR #82, opened 2023-03-31
Feature/sdk - signAndSendTransaction

tanlethanh:


.gitignore
```diff
@@ -56,4 +56,5 @@ yarn-error.log*
 # vscode
 .vscode
 .prettierrc
-.history
\ No newline at end of file
+.history
+sample
```
fluctlight-kayaba: no change at all, should keep last line empty as it is

tanlethanh: This problem has been resolved

fluctlight-kayaba (approved):

PR #90, opened 2023-04-05
Screen/explore profile

dev-tronghieu:


apps/web/src/screens/Dashboard/index.tsx
```diff
@@ -1,10 +1,15 @@
...
+import DashboardLayout from './DashboardLayout';
 
 export const DashboardScreen: FC = () => {
 	return (
 		<Stack>
-			<Text>DashboardScreen</Text>
+			<DashboardLayout>
+				<Outlet />
+			</DashboardLayout>
 		</Stack>
```
FME849, line 13: is it possible to remove outside unused `Stack` and bring out `DashboardLayout`?

apps/web/src/screens/Explore/LayoutCard.tsx
FME849, line 1: consider this is a component, imo, it should be grouped with other components and do not put it together with screens

apps/web/src/screens/Explore/LayoutCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+	id: string;
+	name: string;
+	description: string;
+	thumbnail: string;
+	logo: string;
+	loveCount: number;
+	activeUsers: number;
+}
+
+const LayoutCard: FC<LayoutCardProps> = (props) => {
```
FME849, line 14: try destructing `props` object

apps/web/src/screens/Explore/LayoutCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+				borderTopLeftRadius={12}
+				borderTopRightRadius={12}
+			/>
+
+			<Stack
+				marginHorizontal={12}
+				marginVertical={16}
+				display="flex"
+				y={-36}
+				zIndex={1}
```
FME849, line 35: is it necessary to use zIndex here?

apps/web/src/screens/Explore/SearchBar.tsx
FME849, line 1: this is also component like `LayoutCard`

apps/web/src/screens/Explore/mock.ts
FME849, line 1: consider change this file name to `internal`. Later, it can be used to store code that support the screen, including mock data

apps/web/src/screens/Profile/Collectibles/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+				</Text>
+				<Text
+					color="#566674"
+					fontSize={12}
+					paddingVertical={4}
+					paddingHorizontal={8}
+					borderWidth={1}
+					borderColor="#56667433"
+					borderRadius={8}
+					onPress={() => console.log('See All')}
```
FME849, line 26: try to move logic outside of UI

apps/web/src/screens/Profile/MainFeatures/FeatureButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+		<Stack display="flex" alignItems="center" gap={6}>
+			<Stack
+				width={50}
+				height={50}
+				backgroundColor="#0694D3"
+				borderRadius={16}
+				display="flex"
+				justifyContent="center"
+				alignItems="center"
+				onPress={onClick}
```
FME849, line 21: prefer using onPress to synchronize with `onPress` prop

apps/web/src/screens/Profile/TokenValue.tsx
```diff
@@ -0,0 +1,41 @@
...
+					justifyContent="center"
+				>
+					<ChevronRight size={12} />
+				</Stack>
+			</Stack>
+			<Stack display="flex" flexDirection="row" alignItems="center" gap={5}>
+				<Text fontSize={40} fontWeight="500" lineHeight={26}>
+					{hideTokenValue ? '****' : '$350,270'}
+				</Text>
+				<Stack onPress={() => setHideTokenValue(!hideTokenValue)}>
```
FME849, line 29: again, if function do not depend on `onPress` event, try to separate it from the UI

apps/web/src/state/app/internal.ts
FME849, line 1: do not commit this file

apps/web/src/screens/Explore/LayoutCard.tsx
dev-tronghieu: zIndex must be used here, otherwise the thumbnail will overlap the logo

apps/web/src/screens/Explore/LayoutCard.tsx
dev-tronghieu: In my POV, we have only one page is the `index.tsx`, all other `*.tsx` are the sub-components of that page. All the sub-components will be grouped in the `index.tsx`. Is this implementation good? Cuz I don't want my `index.tsx` has so much code inside it.

apps/web/src/screens/Explore/LayoutCard.tsx
dev-tronghieu: I was thinking about using destructuring as any other components, since this one has lots of attributes (7 in total, maybe more in the future). That's why I think passing an object `props` with be cleaner.
How do you think? Should I change it back to the destructuring syntax?

apps/web/src/screens/Dashboard/index.tsx
dev-tronghieu: Nice point, I forgot to remove it lol xD

apps/web/src/state/app/internal.ts
dev-tronghieu: lemme revert it

apps/web/src/screens/Explore/LayoutCard.tsx
FME849: I mean you can move components to `screens/Explore/components/LayoutCard.tsx`, now it stays in the same level as `index.tsx`

apps/web/src/screens/Explore/SearchBar.tsx
dev-tronghieu: Same idea as the above comment

apps/web/src/screens/Explore/LayoutCard.tsx
FME849: I prefer destructuring

apps/web/src/screens/Profile/Collectibles/index.tsx
dev-tronghieu: Like this?

`const handlePress = () => console.log('See All');`

Then

`onPress={handlePress}`

Or I must separate logic to the parent which invokes that component?

apps/web/src/screens/Explore/LayoutCard.tsx
dev-tronghieu: Got it xD

apps/web/src/screens/Profile/Collectibles/index.tsx
FME849: yes I mean this `onPress={handlePress}`

apps/web/src/screens/Explore/LayoutCard.tsx
FME849: okay let it be for now, will check again with @fluctlight-kayaba

PR #131, opened 2023-04-20
Receive token screen

ltminhthu:


apps/web/src/screens/Profile/modals/ReceiveToken/components/WalletAddress.tsx
```diff
@@ -0,0 +1,67 @@
...
+			bindingDirection: BindDirections.InnerTop,
+			positionOffset: { x: 0, y: 20 },
+			withoutMask: true,
+		});
+		setTimeout(() => {
+			modalActions.hide('copied_announcement');
+		}, 1000);
+	};
+	return (
+		<Button
```
FME849, line 29: separate `return` with previous section by 1 empty line

apps/web/src/screens/Profile/modals/ReceiveToken/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+import ModalHeader from '../components/ModalHeader';
+import ModalWrapper from '../components/ModalWrapper';
+
+import QRCode from './components/QRCode';
+import WalletAddress from './components/WalletAddress';
+
+const ReceiveTokenScreen: FC<{ config: ModalConfigs }> = ({ config }) => {
+	return (
+		<ModalWrapper>
+			<Stack minHeight="70vh">
```
FME849, line 14: avoid using `vh` and `vw` because these units work unpredictably on mobile

apps/web/src/screens/Profile/modals/components/ModalHeader.tsx
```diff
@@ -0,0 +1,35 @@
...
+				{content}
+			</Text>
+			<Stack
+				onPress={() => {
+					modalActions.destroy(config.id);
+				}}
+				cursor="pointer"
+			>
+				<Times size={16} />
+			</Stack>
```
FME849, line 30: why not use `Button` here? Beside `cursor` that indicate clickable button, the feedback on interaction like reducing opacity is also required to provide a better user experience

apps/web/src/screens/Profile/modals/ReceiveToken/index.tsx
ltminhthu: what should i use instead of "vh" in this case?

ltminhthu: > I guess paddingTop or paddingSomething with fixed value like `60` if we just want to separate the Modal from screen edges. Not sure I understand the situation hehe.

i got u, sir, thank uuu

FME849 (approved):

PR #138, opened 2023-04-27
add copied popup 

FME849:


apps/web/src/screens/Setting/components/MyWallets/Wallet.tsx
```diff
@@ -1,7 +1,16 @@
-import { FC } from 'react';
+import { type FC } from 'react';
+import { Image, StyleSheet } from 'react-native';
+import CopiedNotification from '@walless/app/components/CopiedNotification';
```
fluctlight-kayaba, line 3: 1. import from `@walless/app` only, should export the Component in `index.ts` file of the package, and then import it. 
2. Make `CopiedNotification` a generic component that not only handle "copied" message.

apps/web/src/screens/Setting/components/MyWallets/Wallet.tsx
```diff
@@ -1,45 +1,82 @@
...
+			maskActiveOpacity: 0,
+			context: {
+				prefix: StyledCopy,
+				message: 'Copied',
+			},
+		});
+
+		setTimeout(() => {
+			modalActions.hide('copied_announcement');
+		}, 1000);
```
fluctlight-kayaba, line 42: off-load this method from Presentation layer, should be a shared Valtio action instead, what we do is to call it here instead.

packages/app/features/NetworkLayout/WalletCard/Address.tsx
```diff
@@ -31,6 +39,24 @@ export const WalletAddress: FC<Props> = ({ index, skin, token }) => {
...
+			context: {
+				prefix: StyledCopy,
+				message: 'Copied',
+			},
+		});
+
+		setTimeout(() => {
+			modalActions.hide('copied_announcement');
+		}, 1000);
 	};
```
fluctlight-kayaba, line 60: like the above, this should be in Valtio action.

apps/web/src/screens/Setting/components/MyWallets/Wallet.tsx
```diff
@@ -1,45 +1,82 @@
...
 
 interface Props {
 	item: PublicKeyDocument;
 	index: number;
 }
 
-const Wallet: FC<Props> = ({ item, index }) => {
+export const Wallet: FC<Props> = ({ item, index }) => {
 	const network = getNetworkInfo(item.network);
+	const onCopy = () => {
```
fluctlight-kayaba, line 23: can use shared Valtio action I guess?

packages/app/features/NetworkLayout/WalletCard/Address.tsx
```diff
@@ -31,6 +37,8 @@ export const WalletAddress: FC<Props> = ({ index, skin, token }) => {
 
 	const onCopy = () => {
```
fluctlight-kayaba, line 38: make a Valtio action for this, `navigator.clipboard.writeText` could become more complicated (because permission) to handle, put it under unified place reduce fragment.

packages/app/components/Notification/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { type FC } from 'react';
```
fluctlight-kayaba, line 1: put this file, and other "modal" type of component - under `packages/app/modals`

fluctlight-kayaba: and name it use convention: `NotificationModal` instead of just Notification (like `SomethingFeature` for feature).

fluctlight-kayaba (approved):

PR #144, opened 2023-04-30
Add search with some fixing

tanlethanh:


apps/web/src/utils/hooks/debounced.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react';
+
+export const useDebounce = (value: unknown, delay: number) => {
+	const [debounced, setDebounced] = useState(value);
+
+	useEffect(() => {
```
fluctlight-kayaba, line 6: not yet debounced, this code will actually run `setDebounced` all the time with delay. Run`clearTimeout` on effect cleanup not yet enough, will need to run in `clearTimeout` every single dependency changes also.

What I thought, just for idea.. could be in-accurate:
```typescript 
const handler = useRef();

useEffect(() => {
  clearTimeout(handler.current); // cancel any timeout if it is pending/existed
  handler.current = setTimeout(() => {...}, delay);
  
  return () => clearTimeout(handler.current);
}, [value, delay]);
```

packages/gui/components/Select/Dropdown.tsx
```diff
@@ -18,6 +20,17 @@ const Dropdown: FC<Props> = ({ config }) => {
 	const { selected, items, title, getRequiredFields, onSelect } =
 		config.context as SelectionContext<object>;
 
+	const [searchText, setSearchText] = useState('');
```
fluctlight-kayaba, line 23: not efficient yet, use `debounce` (for function instead of state) from lodash and then `useMemo` to memorize filtered item.

packages/gui/components/Select/Dropdown.tsx
```diff
@@ -28,26 +41,74 @@ const Dropdown: FC<Props> = ({ config }) => {
...
+				style={[
+					mutualStyles.button,
+					mutualStyles.focus,
+					styles.dropdownContainer,
+				]}
+			>
+				<View style={[mutualStyles.button, mutualStyles.focus, styles.input]}>
+					<TextInput
+						placeholder="Search"
+						onChangeText={(text) => setSearchText(text)}
```
fluctlight-kayaba, line 62: use lodash's debounce here: e.g `onChangeText={debouncedChangeText}`

fluctlight-kayaba: could be `lodash throttle` instead of `lodash debounce`, the idea is to limit too frequent function calls in short period of time.

packages/gui/components/Select/Dropdown.tsx
```diff
@@ -18,6 +20,17 @@ const Dropdown: FC<Props> = ({ config }) => {
 	const { selected, items, title, getRequiredFields, onSelect } =
 		config.context as SelectionContext<object>;
 
+	const [searchText, setSearchText] = useState('');
+
+	const debouncedSearchText = useDebounce(searchText, 600);
+
+	const renderedItems = items.filter((item) => {
```
fluctlight-kayaba, line 27: with lodash `throttle` or `debounce`, that debounced method just internally `setSearchText` like normal.

this place should be something like:
```typescript
const filteredItems = useMemo(() => {
  return items.filter(() => {
     // searchText always up-to-date inside this memo, use it to filter out data
  });
}, [searchText]);
```

packages/gui/hooks/debounced.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react';
```
fluctlight-kayaba, line 1: `1.` I as commented above, we may not need this 
`2.` Duplicated code, if we have more than 1 place needed something. Put it under `package/app` or `package/gui` and then other place use from there.

packages/gui/index.ts
```diff
@@ -6,6 +6,7 @@ export * from './components/Select';
 export * from './components/Slider';
 export * from './components/Text';
 export * from './components/View';
+export * from './hooks';
```
fluctlight-kayaba, line 9: anyway, I think hooks is more related to the domain of `@walless/app` instead of `@walless/gui`.

apps/web/src/utils/hooks/debounced.ts
tanlethanh: Yes Sir! I realized this last night and I will refactor this hook to work effeciently

PR #148, opened 2023-05-06
refactor Solana and Sui layout to display token list 

FME849:


packages/app/state/tokenList.ts
```diff
@@ -0,0 +1,10 @@
+import { type TokenDocument } from '@walless/store';
```
fluctlight-kayaba, line 1: we dont need this state fragment? we already have useTokens() hook and state associate to it.

FME849: already removed fragment state and change it to `unknown`

PR #151, opened 2023-05-07
Add Input gui and add Select styles

tanlethanh:


packages/gui/components/Input/index.tsx
```diff
@@ -0,0 +1,85 @@
+import { FC, ReactNode, useState } from 'react';
```
fluctlight-kayaba, line 1: nice to have: 
```typescript
import { type FC, type ReactNode } from 'react';
```

fluctlight-kayaba (approved):

PR #169, opened 2023-05-14
update new UI for remove layout

ltminhthu:


packages/store/plugins/delete.ts
```diff
@@ -0,0 +1,8 @@
+export async function docRemove<T extends object>(
```
fluctlight-kayaba, line 1: use name: `remove` if possible

apps/web/src/screens/Explore/components/LayoutCard/index.tsx
```diff
@@ -1,15 +1,30 @@
...
+	onLovePress,
+	onAddPress,
+	isAdded,
+}) => {
 	const { storeMeta } = item;
 	const coverSrc = { uri: storeMeta.coverUri };
 	const iconSrc = { uri: storeMeta.iconUri };
 
+	const handleRemoveLayout = async () => {
+		const doc = await modules.storage.get(item._id);
```
fluctlight-kayaba, line 22: use `storage.remove` plugin added in Pouch instance

packages/store/plugins/delete.ts
ltminhthu: i causes error, sir

fluctlight-kayaba (approved):

PR #178, opened 2023-05-23
add and use private setting from DB

FME849:


apps/web/src/screens/Dashboard/Solana.tsx
```diff
@@ -15,7 +15,9 @@ import { Stack } from '@walless/ui';
 import { layoutTabs } from 'screens/Dashboard/shared';
 import { appActions } from 'state/app';
 import { showReceiveModal } from 'state/app/modal';
-import { usePublicKeys, useTokens } from 'utils/hooks';
+import { settingsActions } from 'state/settings';
+import { settingState } from 'state/settings/internal';
```
tanlethanh, line 19: I thinks these objects should be exported in state/settings `settingState`

apps/web/src/screens/Dashboard/Solana.tsx
```diff
@@ -26,6 +28,7 @@ interface Props {
 
 export const SolanaDashboard: FC<Props> = () => {
 	const [activeTabIndex, setActiveTabIndex] = useState(0);
+	const { _id, isPrivate } = useSnapshot(settingState);
```
tanlethanh, line 31: Maybe we need another name for `isPrivate`. When we have multiple settings, it causes confusion what need to private.

apps/web/src/screens/Dashboard/Solana.tsx
```diff
@@ -56,6 +59,10 @@ export const SolanaDashboard: FC<Props> = () => {
 		appActions.showSendModal(Networks.solana);
 	};
 
+	const handleChangePrivateSetting = (next: boolean) => {
+		settingsActions.updatePrivateSettings({ _id: _id || '', isPrivate: next });
```
tanlethanh, line 63: like above, need another name for `isPrivate`

apps/web/src/screens/Dashboard/Sui.tsx
```diff
@@ -14,7 +14,9 @@ import { Copy } from '@walless/icons';
 import { Stack } from '@walless/ui';
 import { appActions } from 'state/app';
 import { showReceiveModal } from 'state/app/modal';
-import { usePublicKeys, useTokens } from 'utils/hooks';
+import { settingsActions } from 'state/settings';
+import { settingState } from 'state/settings/internal';
```
tanlethanh, line 18: It's the same with the `Dashboard/Solana` comment

apps/web/src/screens/Profile/components/TokenValue.tsx
```diff
@@ -1,12 +1,18 @@
...
+import { Hoverable } from '@walless/gui';
 import { Eye, EyeOff } from '@walless/icons';
 import { Stack, Text } from '@walless/ui';
+import { settingsActions } from 'state/settings';
+import { settingState } from 'state/settings/internal';
+import { useSnapshot } from 'valtio';
 
 const TokenValue = () => {
-	const [hideTokenValue, setHideTokenValue] = useState(true);
+	const { _id, isPrivate } = useSnapshot(settingState);
```
tanlethanh, line 9: `hideTokenValue` may have more meaning, I think ~~~

apps/web/src/state/settings/internal.ts
```diff
@@ -0,0 +1,9 @@
+import { proxy } from 'valtio';
+
+export interface SettingState {
+	_id?: string;
+	_rev?: string;
+	isPrivate?: boolean;
+}
+
+export const settingState = proxy<SettingState>({});
```
tanlethanh, line 9: It may be great to have a default setting, we don't need to handle undefined problems.

packages/app/features/NetworkLayout/WalletCard/index.tsx
```diff
@@ -19,17 +19,20 @@ interface Props {
 	index?: number;
 	item: PublicKeyDocument;
 	skin: CardSkin;
+	isPrivate: boolean;
 	onCopyAddress?: (value: string) => void;
+	onChangePrivateSetting?: (value: boolean) => void;
```
tanlethanh, line 24: What private is? the same above problem.

apps/web/src/screens/Dashboard/Sui.tsx
```diff
@@ -56,6 +59,10 @@ export const SuiDashboard: FC<Props> = () => {
 		appActions.showSendModal(Networks.sui);
 	};
 
+	const handleChangePrivateSetting = (next: boolean) => {
+		settingsActions.updatePrivateSettings({ _id: _id || '', isPrivate: next });
```
tanlethanh, line 63: Use a default setting can make it clear `{ _id }`

apps/web/src/screens/Dashboard/Sui.tsx
```diff
@@ -66,7 +73,9 @@ export const SuiDashboard: FC<Props> = () => {
 							index={index}
 							item={item}
 							skin={suiCardSkin}
+							isPrivate={isPrivate || false}
```
tanlethanh, line 76: The same as the above comment

apps/web/src/state/settings/index.ts
```diff
@@ -0,0 +1,27 @@
...
+import { type SettingDocument } from '@walless/store';
+
+import { settingState } from './internal';
+
+export const settingsActions = {
+	setSettings: (settings: SettingDocument) => {
+		if (settings?._id) {
+			settingState._id = settings._id;
+			settingState._rev = settings._rev;
+			settingState.isPrivate = settings.isPrivate;
```
tanlethanh, line 11: `isPrivate`

apps/web/src/state/settings/index.ts
```diff
@@ -0,0 +1,27 @@
...
+		}
+	},
+	updatePrivateSettings: async ({
+		_id,
+		isPrivate,
+	}: {
+		_id: string;
+		isPrivate: boolean;
```
tanlethanh, line 19: Hehe

apps/web/src/screens/Dashboard/Solana.tsx
FME849: I replaced `isPrivate` by `hideBalance`

apps/web/src/screens/Dashboard/Solana.tsx
```diff
@@ -26,6 +27,7 @@ interface Props {
 
 export const SolanaDashboard: FC<Props> = () => {
 	const [activeTabIndex, setActiveTabIndex] = useState(0);
+	const { _id, hideBalance } = useSnapshot(settingState);
```
fluctlight-kayaba, line 30: because we have more than 1 place using this, will be good to make a custom hook named `useSetting` which:
- return `setting` object, including `hideBalance`
- return `updateSetting` method (created by `useCallback`)
- return other useful thing in the future

With that `useSetting` hook above, we can:
```typescript
const { setting, setSetting } = useSetting();

setting.hideBalance; // live update once something change from the db
setSetting({ hideBalance: true }); // created by `useCallback` in sync with custom-hook's dependency
```

apps/web/src/screens/Dashboard/Sui.tsx
```diff
@@ -26,6 +27,7 @@ interface Props {
 
 export const SuiDashboard: FC<Props> = () => {
 	const [activeTabIndex, setActiveTabIndex] = useState(0);
+	const { _id, hideBalance } = useSnapshot(settingState);
```
fluctlight-kayaba, line 30: promote this as `useSetting`, see above comment.

apps/web/src/screens/Profile/components/TokenValue.tsx
```diff
@@ -1,12 +1,17 @@
-import { useState } from 'react';
+import { Hoverable } from '@walless/gui';
 import { Eye, EyeOff } from '@walless/icons';
 import { Stack, Text } from '@walless/ui';
+import { settingsActions, settingState } from 'state/settings';
+import { useSnapshot } from 'valtio';
 
 const TokenValue = () => {
-	const [hideTokenValue, setHideTokenValue] = useState(true);
+	const { _id, hideBalance } = useSnapshot(settingState);
```
fluctlight-kayaba, line 8: need `useSetting` effect

apps/web/src/screens/Profile/components/TokenValue.tsx
```diff
@@ -1,12 +1,17 @@
...
 
-	const handleToggleTokenValue = () => {
-		setHideTokenValue(!hideTokenValue);
+	const handleToggleTokenValue = async () => {
```
fluctlight-kayaba, line 10: `useSetting`

apps/web/src/state/live/index.ts
```diff
@@ -22,10 +24,13 @@ export const initializeLiveState = async () => {
 	const publicKeys = publicKeyResponse.docs as PublicKeyDocument[];
 	const tokenResponse = await modules.storage.find(selectors.allTokens);
 	const tokens = tokenResponse.docs as TokenDocument[];
+	const settingsResponse = await modules.storage.find(selectors.settings);
```
fluctlight-kayaba, line 27: don't use `find`, use `modules.storage.get()` we need one single item, and we do know the id of it.

apps/web/src/screens/Dashboard/Sui.tsx
```diff
@@ -56,6 +58,13 @@ export const SuiDashboard: FC<Props> = () => {
 		appActions.showSendModal(Networks.sui);
 	};
 
+	const handleChangePrivateSetting = (next: boolean) => {
+		settingsActions.updateHiddenBalanceSettings({
+			_id,
+			hideBalance: next,
+		});
```
fluctlight-kayaba, line 65: we already now `_id` of the setting, don't ask for it again.
also, the name of the function quite confusing.

we can just use `settingsActions.setPrivacy(value: bool);`

apps/web/src/state/settings/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { settingState } from './internal';
+
+export const settingsActions = {
+	setSettings: (settings: SettingDocument) => {
+		if (settings._id) {
+			settingState._id = settings._id;
+			settingState.hideBalance = settings.hideBalance;
+		}
+	},
+	updateHiddenBalanceSettings: async ({
```
fluctlight-kayaba, line 13: refactor this function, we can chat for further detail.

apps/web/src/state/settings/index.ts
```diff
@@ -0,0 +1,28 @@
...
+		hideBalance,
+	}: {
+		_id: string;
+		hideBalance: boolean;
+	}) => {
+		await modules.storage.upsert<SettingDocument>(_id, async (doc) => {
+			doc.hideBalance = hideBalance;
+
+			return doc;
+		});
```
fluctlight-kayaba, line 24: once the user signed-in this setting Document must be ready there, no `upsert` is or `put` needed, I think we don't need this method at all.

apps/web/src/state/settings/internal.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export interface SettingState {
+	_id: string;
+	hideBalance: boolean;
+}
+
+export const settingState = proxy<SettingState>({
+	_id: '',
+	hideBalance: true,
+});
```
fluctlight-kayaba, line 11: - if we have new state, that would be use in Mobile (this state will be use in Mobile) - then it should be in `packages/app` instead
- since we put the setting in existed db object, "settingState" is actually the `appState` we already had, I think we better put this `hideBalance` just in the `appState`
- then, remove this `settingState` I guess

FME849: do you think there will be more settings share among platforms? should we create new `settingsState` instead of putting more settings to `appState`?

fluctlight-kayaba: in that case, it would be an object inside `setting` document (instead of a field like now)

with it, we'll have an object e.g `setting.preferences` (now we have setting.hideBalance), and from within `appState`, we'll have `appState.perferences`.

apps/web/src/state/settings/index.ts
FME849: the requirement is when user open extension again, they see the same private setting as the last time they closed extension. Therefore, I think we need a method to update the setting Document, am I right?

fluctlight-kayaba: oops, yes! you're right, I forgot that `upsert` is for update as well.

apps/web/src/state/live/index.ts
```diff
@@ -22,10 +24,14 @@ export const initializeLiveState = async () => {
 	const publicKeys = publicKeyResponse.docs as PublicKeyDocument[];
 	const tokenResponse = await modules.storage.find(selectors.allTokens);
 	const tokens = tokenResponse.docs as TokenDocument[];
+	const settings = (await modules.storage.safeGet(
+		'settings',
+	)) as SettingDocument;
```
fluctlight-kayaba, line 29: use this: `await modules.storage.safeGet<SettingDocument>('settings');`

packages/store/utils/helper.ts
```diff
@@ -10,4 +10,5 @@ export const selectors = {
 	suiKeys: { selector: { type: 'PublicKey', network: Networks.sui } },
 	allTokens: { selector: { type: 'Token' } },
 	trustedDomains: { selector: { type: 'TrustedDomain' } },
+	settings: { selector: { type: 'Setting' } },
```
fluctlight-kayaba, line 13: remove this

packages/app/state/app.ts
```diff
@@ -13,6 +13,7 @@ export interface AppState {
 	profileReady: boolean;
 	profileLoading: boolean;
 	profile: UserProfile;
+	settingConfig: SettingConfig;
```
fluctlight-kayaba, line 16: `appState.config`

apps/web/src/state/settings/index.ts
```diff
@@ -0,0 +1,20 @@
...
+		}
+	},
+	setPrivacy: async ({ hideBalance }: { hideBalance: boolean }) => {
+		await modules.storage.upsert<SettingDocument>(id, async (doc) => {
+			doc.config.hideBalance = hideBalance;
+
+			return doc;
+		});
+	},
+};
```
fluctlight-kayaba, line 20: we don't want new setting state at all, even if you only define actions for it - define that actions inside appState.

fluctlight-kayaba (approved):
pixel perfect!

PR #183, opened 2023-05-30
navigate to latest screen

ltminhthu:


apps/web/src/screens/Dashboard/index.tsx
```diff
@@ -28,6 +28,8 @@ export const DashboardScreen: FC = () => {
 		await router.navigate('/');
 	};
 
+	localStorage.setItem('pathname', pathname);
+
```
FME849, line 32: I think this feature could be reusable on mobile, thus we should not use `localStorage` here. You could try to store the layout `id` in setting document in DB and then navigate based on that `id`.

ltminhthu: do we need to synchronize the `layout_id` between mobile and extension?

ltminhthu: > I think this feature could be reusable on mobile, thus we should not use `localStorage` here. You could try to store the layout `id` in setting document in DB and then navigate based on that `id`.

btw, do we need to store screen like "setting" or "profile"

FME849: > do we need to synchronize the `layout_id` between mobile and extension?

do you mean synchronize last project `id` between mobile and extension? I think those platforms share the feature not the data

FME849: > > I think this feature could be reusable on mobile, thus we should not use `localStorage` here. You could try to store the layout `id` in setting document in DB and then navigate based on that `id`.
> 
> btw, do we need to store screen like "setting" or "profile"

I can't think of any purpose to save `setting` or `profile` screens, but we could check again with others.

apps/web/src/screens/Dashboard/index.tsx
```diff
@@ -28,6 +29,17 @@ export const DashboardScreen: FC = () => {
...
+		if (pathname !== setting?.currentScreen) {
+			console.log(setting?.currentScreen);
+			await setPathname({ currentScreen: pathname });
+		}
+	};
+
+	useEffect(() => {
+		setLatestPathname();
+	}, [pathname]);
+
```
FME849, line 42: I think `async/await` is not required here, we can reduce these line like this:
```typescript
  useEffect(() => {
    if (pathname !== setting?.currentScreen) {
     setPathname(pathname);
    }
  }, [pathname]);
```

apps/web/src/state/app/settings.ts
```diff
@@ -18,3 +18,15 @@ export const setPrivacy = async ({ hideBalance }: { hideBalance: boolean }) => {
...
+	currentScreen,
+}: {
+	currentScreen: string;
+}) => {
+	await modules.storage.upsert<SettingDocument>(id, async (doc) => {
+		doc.config.currentScreen = currentScreen;
+
+		return doc;
+	});
+};
```
FME849, line 32: I realized that there's a parameter mistake in my previous `setPrivacy` that it does not need to destruct `hideBalance` there. I will explain more in chat.

apps/web/src/screens/Dashboard/index.tsx
```diff
@@ -28,6 +29,17 @@ export const DashboardScreen: FC = () => {
 		await router.navigate('/');
 	};
 
+	const setLatestPathname = async () => {
+		if (pathname !== setting?.currentScreen) {
+			console.log(setting?.currentScreen);
```
fluctlight-kayaba, line 34: should clean console.log before push

apps/web/src/screens/Dashboard/index.tsx
```diff
@@ -28,6 +29,17 @@ export const DashboardScreen: FC = () => {
...
+			await setPathname({ currentScreen: pathname });
+		}
+	};
+
+	useEffect(() => {
```
fluctlight-kayaba, line 39: this observe/watch flow should be in higher level, e.g App to ensure it work for other routes as well

packages/app/state/app.ts
```diff
@@ -25,5 +25,6 @@ export const appState = proxy<AppState>({
 	authenticationLoading: false,
 	config: {
 		hideBalance: true,
+		currentScreen: '/',
```
fluctlight-kayaba, line 28: propose use more generic name e.g `latestLocation`

packages/core/utils/entityTypes.ts
```diff
@@ -9,6 +9,7 @@ export interface Setting {
 
 export interface Config {
 	hideBalance: boolean;
+	currentScreen: string;
```
fluctlight-kayaba, line 12: consider another name

PR #189, opened 2023-06-01
Fix styles and add responsive in `apps/docs`

tanlethanh:


apps/docs/src/components/HomeLayout/TopNavigation/index.tsx
```diff
@@ -41,7 +41,7 @@ export const TopNavigation: FC<Props> = ({ docs, docsTree }) => {
 					);
 				})}
 			</View>
-			<View>
+			<View style={{ width: styles.logo.width, alignItems: 'flex-end' }}>
```
FME849, line 44: we could create a constant width and the styles share it, like this: 
```typescript
const sharedWidth = 256;
const styles = Stylesheet.create({
  logo: {
    width: sharedWidth,
  },
  button: {
    width: sharedWidth,
  },
})
```

tanlethanh: Ok, I got it

apps/docs/src/components/HomeLayout/Content/index.tsx
```diff
@@ -41,7 +36,7 @@ export default Content;
 const styles = StyleSheet.create({
 	container: {
 		paddingTop: 40,
-		gap: 30,
-		alignItems: 'center',
+		maxWidth: 1000,
```
fluctlight-kayaba, line 39: should be something like 1024 or 1280 or 1620 I think, please check which number is that in our Home (Landing page) and use same value here -- should put this `maxWidth` in a reuse-able module like `utils/config.ts`

apps/docs/src/components/HomeLayout/Content/index.tsx
```diff
@@ -41,7 +36,7 @@ export default Content;
 const styles = StyleSheet.create({
 	container: {
```
fluctlight-kayaba, line 37: I think this "container" style could be in `utils/style.ts` like what we did in Landing page, make it reuse-able

apps/docs/src/components/HomeLayout/index.tsx
```diff
@@ -20,11 +21,28 @@ export const HomeLayout: FC<Props> = ({ docs, params, docsTree }) => {
...
 	);
 };
 
+const styles = StyleSheet.create({
+	container: {
+		height: '100vh',
+	},
+	contentContainer: {
+		width: '100%',
+		maxWidth: 1620,
```
fluctlight-kayaba, line 43: make this a constant, e.g in `utils/config.ts` and the reuse it

packages/markdown/rules/codeBlock.ts
```diff
@@ -26,6 +26,7 @@ const codeBlock: ParserRule & ReactOutputRule = {
 						marginTop: '10px',
 					},
 				},
+				fontSize: 14,
```
fluctlight-kayaba, line 29: fontSize 14 should be default tho?

tanlethanh: No, This size comes from default of React Syntax Highlighter package, we use its component

apps/docs/src/components/HomeLayout/index.tsx
```diff
@@ -20,11 +21,28 @@ export const HomeLayout: FC<Props> = ({ docs, params, docsTree }) => {
...
-		</Fragment>
+		<View style={styles.container}>
+			<View style={styles.contentContainer}>
+				<TopNavigation docs={docs} docsTree={docsTree} />
+			</View>
+			<ScrollView showsVerticalScrollIndicator={false}>
+				<View style={styles.contentContainer}>
+					<Content docsTree={docsTree} docs={docs} params={params} />
+				</View>
+			</ScrollView>
```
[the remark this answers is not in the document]
tanlethanh: `ScrollView` is already full-width and centered. But in edge browser, this `indicator` takes width from total width of `ScrollView`. That makes centering is missmatched between `top navigation`  and `content`
I need help to handle this case

apps/docs/src/components/HomeLayout/index.tsx
```diff
@@ -20,11 +22,23 @@ export const HomeLayout: FC<Props> = ({ docs, params, docsTree }) => {
...
-		</Fragment>
+		<View style={styles.container}>
+			<View style={sharedStyles.container}>
+				<TopNavigation docs={docs} docsTree={docsTree} />
+			</View>
+			<ScrollView showsVerticalScrollIndicator={false}>
+				<View style={sharedStyles.container}>
+					<Content docsTree={docsTree} docs={docs} params={params} />
+				</View>
+			</ScrollView>
```
fluctlight-kayaba, line 33: use this to avoid create extra `<View/>` Node: 
```typescript
<ScrollView contentContainerStyle={sharedStyles.container}>
  <Content docsTree={docsTree} docs={docs} params={params} />
</ScrollView>
```

FME849 (approved):

fluctlight-kayaba (approved):

PR #205, opened 2023-06-12
Fix UI and refactor code

tanlethanh:


apps/docs/src/components/HomeLayout/index.tsx
```diff
@@ -17,29 +20,72 @@ interface Props {
...
+	let scrollPaddingLeft = 20;
+	if (responsiveLevel >= 1) {
+		scrollPaddingLeft = 20;
+	} else if (windowSize.width < 1200) {
+		scrollPaddingLeft = 300;
+	} else if (windowSize.width < 1340) {
+		scrollPaddingLeft = 200;
+	} else if (windowSize.width < 1400) {
+		scrollPaddingLeft = 100;
 	}
```
fluctlight-kayaba, line 42: make a util function right bellow this file, but outside of the Component, maybe next to `const styles` - in case the method could be use in other place, put it under `utils/...`

apps/docs/src/pages/_app.tsx
```diff
@@ -15,12 +19,19 @@ export const App: FC<AppProps> = ({ Component, pageProps }) => {
 		setRender({});
 	}, []);
 
+	useEffect(() => {
+		modalActions.setContainerRef(containerRef);
+	}, []);
+
```
fluctlight-kayaba, line 25: put this together with above empty useEffect[]

apps/docs/src/pages/index.tsx
```diff
@@ -15,7 +15,7 @@ import { type DocsTree } from 'utils/types';
 const AnimatedBox = Animated.createAnimatedComponent(View);
```
fluctlight-kayaba, line 15: use `import { AnimatedView } from '@walless/gui';` instead

apps/docs/src/utils/engine.ts
```diff
@@ -34,18 +35,23 @@ export const loadMarkdown = async (
...
+					node,
 				),
 			),
 		);
-
+		node.children = children.flatMap((child) => (child ? child : []));
 		node.children.sort(compareDocsTree);
 	} else {
-		node.content = fs.readFileSync(markdownPath, 'utf-8');
+		console.log(fileName);
```
fluctlight-kayaba, line 50: remove this line

fluctlight-kayaba (approved):

PR #209, opened 2023-06-14
Wallet standard - Review code

tanlethanh:


apps/web/scripts/kernel/handlers/connect.ts
```diff
@@ -29,16 +32,45 @@ export const handleConnect: MessengerCallback = async (payload, channel) => {
...
+export const handleResolveConnect: MessengerCallback = async (payload) => {
+	const { requestId = '' } = payload;
+	const requestSource = requestMap[requestId];
+	if (!requestSource?.resolve) {
+		if (!payload.isApproved) {
+			requestSource.channel.postMessage({
+				from: 'walless@kernel',
+				requestId,
+				message: Message.REJECT_REQUEST_CONNECT,
+			});
```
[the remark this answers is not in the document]
tanlethanh: I mean we need to create a convention for messaging about success and failed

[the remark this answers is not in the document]
FME849: I also think we need fields like `error` and `responseCode`. Is that what you mean?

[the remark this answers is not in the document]
tanlethanh: I mean in messaging, when receiving a reject, It will raise error. We need an util to handle this (inside messaging engine code) before sdk receive response. @fluctlight-kayaba We need your opinion.

apps/web/scripts/kernel/handlers/kernel.ts
```diff
@@ -1,17 +1,23 @@
...
+		} else if (payload.type === RequestType.RESOLVE_REQUEST_CONNECT) {
+			handleResolveConnect(payload, channel);
 		} else if (payload.type == RequestType.GET_ENDPOINT_ON_SOLANA) {
 			solanaHandler.getEndpoint(payload, channel);
 		} else if (payload.type === RequestType.SIGN_MESSAGE_ON_SOLANA) {
 			solanaHandler.handleSignMessage(payload, channel);
+		} else if (payload.type === RequestType.REQUEST_PAYLOAD) {
+			solanaHandler.handleRequestPayload(payload, channel);
+		} else if (payload.type === RequestType.RESOLVE_REQUEST_SIGNATURE) {
+			solanaHandler.handleResolveRequestSignature(payload, channel);
```
tanlethanh, line 20: Working with sign and send (both transaction and message), I think we need mechanism like `forwarding`. It means before passing to handler, we weed a forwarder to decide next action (like open popup). This mechanism will detach the handler at kernel with comunication between `sdk`, `kernel` and `ui`

apps/web/scripts/kernel/handlers/shared.ts
```diff
@@ -0,0 +1,30 @@
...
+import { type MessagePayload, type PopupType } from '@walless/messaging';
+
+interface RequestContent {
+	channel: MiniBroadcast;
+	popup?: chrome.windows.Window;
+	payload: MessagePayload;
+	resolve: boolean;
+}
+
+export const requestMap: Record<string, RequestContent> = {};
```
tanlethanh, line 11: I think we need to make it universal, like request pool, and when we want to response instead of using `channel.postMessge` and include `requestId` inside payload, we will use another method like `messaging.response({ to: requestId, payload})`.

apps/web/scripts/kernel/handlers/solanaHandler.ts
```diff
@@ -37,47 +38,126 @@ export const handleSignMessage: MessengerCallback = async (
...
+	requestMap[requestId] = {
+		channel,
+		payload,
+		popup,
+		resolve: false,
+	};
 };
 
-export const handleSignTransaction: MessengerCallback = async (
+export const handleResolveRequestSignature: MessengerCallback = async (
```
tanlethanh, line 51: `signature` is just the thing `kernel` must response to `sdk`, under some action to generate this signature. The name could be meaningful

apps/web/scripts/kernel/handlers/solanaHandler.ts
```diff
@@ -37,47 +38,126 @@ export const handleSignMessage: MessengerCallback = async (
...
+			requestId: requestSource.payload.requestId,
+			message,
+		});
+
+		channel.postMessage({
+			from: 'walless@kernel',
+			requestId: payload.requestId,
+			responseCode: ResponseCode.ERROR,
+			message: 'Request rejected',
+		});
```
tanlethanh, line 73: Continue to previous comment, I think it might be refactored like:
```typescript
messaging.repsonse({ 
          to: requestId, 
          responseCode: ResponseCode.ERROR, 
          payload 
})
```
- `to` and `responseCode` is required, and `from` will be attached from var which is initalized with messaging
- this `response` method will look at message pool, then we can pass over multiple channels without messy and conflict.
@FME849 @fluctlight-kayaba I mean we need to redifine this

tanlethanh (commented):
image
Make it reusable and clean

apps/docs/src/utils/module.d.ts
fluctlight-kayaba, line 1: should not be here, belongs to another commits/branch

apps/docs/tsconfig.json
fluctlight-kayaba, line 1: should not be here, belongs to another commits/branch

apps/landing/tsconfig.json
fluctlight-kayaba, line 1: maybe should not be here, belongs to another commits/branch

apps/landing/utils/modules.d.ts
fluctlight-kayaba, line 1: should not be here, belongs to another commits/branch

apps/web/scripts/kernel/handlers/connect.ts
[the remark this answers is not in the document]
fluctlight-kayaba: I have no clear clue if Messaging should/could handle that, initial idea is to decouple child modules, it should cover thing that generic enough. While error handing may different between different parts (e.g sdk vs content-script), keep in mind that sdk don't use messing module at all.

While we need unified convention/util to handle success/failure, maybe it won't be part of the messaging package, I don't know, @tanlethanh if you already have direction/idea just go  ahead.

apps/web/scripts/kernel/handlers/kernel.ts
fluctlight-kayaba: Go a head if you have concrete idea, will be useful to provide code snippet or discuss offline, I don't really understand what you're mentioning.

apps/web/scripts/kernel/handlers/shared.ts
fluctlight-kayaba: I think that `requestMap` currently act as the request pool.

fluctlight-kayaba: we do have more than one place using this pattern, I'm not sure unified them benefits us enough, more than the effort it takes.. but I do love to see alternative implementations.

packages/network/src/solana/transactions.ts
```diff
@@ -47,3 +48,13 @@ export const signAndSendTransaction = async (
...
 };
+
+export const simulateTransaction = async (
+	connection: Connection,
+	transaction: string,
+) => {
+	const tx = VersionedTransaction.deserialize(decode(transaction));
+
+	const simulatedTx = await connection.simulateTransaction(tx);
+	console.log(simulatedTx, '<-----');
```
fluctlight-kayaba, line 59: remove this

packages/store/plugins/delete.ts
fluctlight-kayaba, line 1: should not be here, commit from another branch

PR #210, opened 2023-06-15
Wl-118 redirect to latest screen after logging in

ltminhthu:


apps/web/src/screens/Dashboard/index.tsx
```diff
@@ -28,6 +29,12 @@ export const DashboardScreen: FC = () => {
 		await router.navigate('/');
 	};
 
+	useEffect(() => {
+		if (pathname !== setting?.latestLocation) {
+			setPathname(pathname);
+		}
+	}, [pathname]);
+
```
FME849, line 37: we already agreed to move `setPathname` outside `DashboardScreen` in the last discussion. However, is that solution possible?

ltminhthu: > we already agreed to move `setPathname` outside `DashboardScreen` in the last discussion. However, is that solution possible?

We discussed about it and all of us agree to put it in `DashboardScreen`

fluctlight-kayaba (approved):

PR #212, opened 2023-06-15
Wl-122 log out and Wl-121 feedback button

ltminhthu:


apps/web/src/screens/Setting/components/index.ts
```diff
@@ -0,0 +1,7 @@
+export { Delimiter } from './Delimiter';
```
fluctlight-kayaba, line 1: if those component could be reuse in other place like Mobile, then it should be in `packages/app`.

apps/web/src/screens/Setting/index.tsx
```diff
@@ -1,12 +1,15 @@
...
-import HelpCenter from './components/HelpCenter';
-import MyWallets from './components/MyWallets';
+import {
+	Delimiter,
+	FollowUs,
+	Header,
+	HelpCenter,
+	LogOut,
+	MyWallets,
+} from './components';
```
fluctlight-kayaba, line 12: prefer `import from './components/ComponentName';` than this, re-export using `index.ts` file may cause unexpected circular deps.

apps/web/src/screens/Setting/index.tsx
```diff
@@ -17,7 +20,13 @@ const SettingScreen = () => {
 			<Delimiter />
 			<Stack gap={16}>
 				<MyWallets />
-				<HelpCenter />
+				<Stack>
```
fluctlight-kayaba, line 23: try to remove Tamagui component (Stack) as long as we face it.

packages/store/plugins/clear.ts
```diff
@@ -0,0 +1,6 @@
+export async function clear(this: PouchDB.Database) {
```
fluctlight-kayaba, line 1: `clear` is not enough to explains, suggest e.g `clearAllDocs`.

packages/store/plugins/index.ts
```diff
@@ -8,4 +9,5 @@ export default {
 	upsert,
 	safeGet,
 	docRemove,
```
fluctlight-kayaba, line 11: if possible, rename this into `removeDoc`

fluctlight-kayaba (approved):

PR #221, opened 2023-06-21
resolve undefined latestPathname bug

dev-tronghieu:


apps/web/src/state/app/settings.ts
```diff
@@ -21,7 +21,8 @@ export const setPrivacy = async (hideBalance: boolean) => {
 
 export const setPathname = async (latestScreen: string) => {
 	await modules.storage.upsert<SettingDocument>(id, async (doc) => {
-		doc.config.latestLocation = latestScreen;
+		if (doc.config.latestLocation !== latestScreen)
+			doc.config.latestLocation = latestScreen;
```
fluctlight-kayaba, line 25: seems not related to `undefined` stuff

dev-tronghieu: With the accounts using the old database, there is no field name `latestLocation` in `doc.config` which causes `undefined latestLocation` error

PR #227, opened 2023-06-28
Wl 116 dapps

dev-tronghieu:


apps/dappstore/utils/module.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'react-native-syntax-highlighter';
+declare module 'react-syntax-highlighter/styles/hljs';
```
tanlethanh, line 2: these modules are unused in this project

apps/dappstore/components/layouts/HomeLayout.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const HomeLayout: FC<Props> = ({ children }) => {
+	return (
+		<View style={sharedStyles.minScreen}>
+			<View style={sharedStyles.container}>
+				<Header />
+			</View>
+			<ScrollView
+				contentContainerStyle={sharedStyles.contentContainer}
+				showsVerticalScrollIndicator={false}
```
tanlethanh, line 20: style the scroll indicator instead of hiding it

tanlethanh (commented):
I think moving all source code into the `apps/dappstore/src` directory might make the source base cleaner. It will separate code and configurations.

apps/dappstore/components/layouts/HomeLayout.tsx
dev-tronghieu: I have contacted Ms. Ngoc for the new UI. Will update this soon

tanlethanh (approved):
@dev-tronghieu I think we could close this PR, resolve confict and move forward. Another update might be in the next PR

PR #229, opened 2023-06-28
use seed phrase module to derive key

tanlethanh:


apps/web/src/state/app/authentication.ts
```diff
@@ -188,16 +198,121 @@ export const storeAuthenticatedRecords = async (
 	login?: UserCredential,
 ): Promise<void> => {
 	await key.reconstructKey();
-	const privateKeys = await key.modules.privateKeyModule.getPrivateKeys();
 
 	if (login?.user) {
 		await setProfile(makeProfile(login));
 	}
 
+	// await initByPrivateKeyModule(passcode);
```
FME849, line 206: give explanation why this comment should be kept

apps/web/src/state/app/authentication.ts
```diff
@@ -188,16 +198,121 @@ export const storeAuthenticatedRecords = async (
...
+			network: Networks.solana,
+		},
+		{
+			path: "44'/784'",
+			network: Networks.sui,
+		},
+		{
+			path: "44'/1729'",
+			network: Networks.tezos,
+		},
```
FME849, line 226: maybe a comment which explain where these `path` come from and/or how to find them in docs would be appreciated

FME849 (approved):

apps/web/src/state/app/authentication.ts
tanlethanh: these paths are default paths for each network, we can find them in the dev docs of each network. I also noted this in [notion task](https://www.notion.so/stormgate/2a39a522cbd443a59ee2eb06ebb9dba2?v=1904b777f28648f28d21db626f9bfec4&p=83aad874791445d880937c1bee6a41ea&pm=s)

packages/core/utils/commonTypes.ts
```diff
@@ -50,7 +50,6 @@ export enum Networks {
 	ethereum = 'ethereum',
 	solana = 'solana',
 	sui = 'sui',
-	tezos = 'tezos',
```
fluctlight-kayaba, line 53: still need this I guess

tanlethanh: Have been duplicated above

PR #241, opened 2023-07-11
nft management and transfer

tanlethanh:


packages/core/utils/entityTypes.ts
```diff
@@ -56,9 +57,14 @@ export interface Token {
...
+	network: string;
+	metadata?: AssetMetadata;
+}
+
 export interface Collectible {
 	network: string;
-	collectionId?: string;
+	collectionId: string;
 	metadata?: AssetMetadata;
 }
```
tanlethanh, line 69: @ltminhthu here is type shape of collectible (aka nft) which includes metadata field as same as token type. In `send feature` we mostly use this field to render. It might be helpful to notice.

packages/engine/solana/crawler.ts
```diff
@@ -25,33 +28,58 @@ export const solanaEngineRunner: EngineRunner<Connection> = {
...
+		promises.push(
+			Promise.all(collectiblePromises).then(async (results) => {
+				collectibleActions.setCollections(
+					flatten(results.map((result) => result.collections)),
+				);
+				collectibleActions.setCollectibles(
+					flatten(results.map((result) => result.collectibles)),
+				);
+			}),
+		);
```
tanlethanh, line 80: @ltminhthu here is where we initialize all nfts with their collections. update state inside engine

packages/engine/solana/collectibles.ts
```diff
@@ -0,0 +1,99 @@
...
+			if (nft.collection) {
+				collectionMetadata = (await mpl
+					.nfts()
+					.findByMint({ mintAddress: nft.collection.address })) as Nft;
+			}
+
+			const collectionId = `${address}/${
+				collectionMetadata?.address.toString() ||
+				nft.collection?.address.toString() ||
+				nft.address.toString()
```
tanlethanh, line 55: in case the NFT is not in any collection, we treat it like a collection including one item

packages/engine/solana/collectibles.ts
```diff
@@ -0,0 +1,99 @@
...
+			}`;
+
+			if (!collections.find((ele) => ele._id === collectionId)) {
```
FME849, line 58: `find` could be replaced by `some` which return `boolean` type and give a more transparent idea about what condition we want to check.

fluctlight-kayaba:
need your help @tanlethanh, resolve conflicts

packages/engine/utils/type.ts
```diff
@@ -14,7 +14,7 @@ export type RunnerContext<T> = NetworkContext<T> & {
 
 export type EngineRunner<T> = {
 	stop: (context: RunnerContext<T>) => Promise<void>;
-	start: (context: RunnerContext<T>) => Promise<void>;
+	start: (context: RunnerContext<T>) => Promise<void | void[]>;
```
fluctlight-kayaba, line 17: better cast `as any` in another end, and keep this as it is (no `void[]`)

fluctlight-kayaba (approved):

PR #249, opened 2023-07-20
feat: wl-191-full-flow-sdk

FME849:


apps/web/scripts/kernel/handlers/common.ts
```diff
@@ -27,9 +31,13 @@ export const handleRequestPayload: HandleMethod = ({
 	responseMethod,
 }) => {
 	const { sourceRequestId, requestId } = payload;
-	const { payload: sourcePayload } = getRequestRecord(sourceRequestId);
+	const { payload: sourcePayload, channel: sourceChannel } =
+		getRequestRecord(sourceRequestId);
 
-	responseMethod(requestId, ResponseCode.SUCCESS, sourcePayload);
+	responseMethod(requestId, ResponseCode.SUCCESS, {
+		...sourceChannel,
```
tanlethanh, line 38: @FME849 can you explain why we need source channel in the response body?

apps/web/scripts/kernel/handlers/common.ts
```diff
@@ -45,3 +53,31 @@ export const handleInstallLayout: HandleMethod = async ({
...
+}) => {
+	const { requestId, id } = payload as InstallLayoutPayload;
+	const isInstalled = await checkInstalledExtensionById(id);
+
+	if (isInstalled) {
+		responseMethod(requestId, ResponseCode.SUCCESS);
+	} else {
+		responseMethod(requestId, ResponseCode.ERROR);
+	}
+};
```
tanlethanh, line 69: For convenience, we might want to make this method support dynamically like if without passing it will return all the installed layouts, users can check it. Or implement another method for it.

apps/web/scripts/kernel/handlers/kernel.ts
```diff
@@ -4,34 +4,39 @@ import { RequestType } from '@walless/messaging';
...
+			handleMethod = commonHandler.handleRequestPayload;
 			requirePrivateKey = false;
 		} else if (type === RequestType.INSTALL_LAYOUT) {
-			handleMethod = handleInstallLayout;
+			handleMethod = commonHandler.handleInstallLayout;
 			requirePrivateKey = false;
+		} else if (type === RequestType.CHECK_INSTALLED_LAYOUT) {
+			handleMethod = commonHandler.handleCheckInstalledLayout;
+			requirePrivateKey = false;
+			requireUserAction = false;
```
tanlethanh, line 35: like previous comment, this request may get all installed layouts, checking existing of a specific layout might be made by user (or from calling side)

apps/web/src/bridge/listeners/index.ts
```diff
@@ -93,6 +93,8 @@ export const getMessageOrTransaction = async (requestId: string) => {
 			};
 		} else if ('transaction' in res) {
 			return res;
+		} else {
+			return res;
 		}
```
tanlethanh, line 98: both of these else statments return `res`

packages/sdk/providers/common.ts
```diff
@@ -3,23 +3,47 @@ import { RequestType } from '@walless/messaging';
...
+export const requestInstallLayout = async (id: string) => {
 	return await sendRequest(
 		{
-			from: 'walless@sdk',
+			from,
 			type: RequestType.INSTALL_LAYOUT,
-			input,
+			id,
+		},
+		30000,
```
tanlethanh, line 26: make this magic number to a global constant with a meaningful name

apps/web/scripts/kernel/handlers/common.ts
FME849: this is the response for `requestPayload` from popup, the source channel include a hidden field called `sender` that has info about dapp (i.e domain, url, title,...)

apps/web/scripts/kernel/handlers/common.ts
FME849: In this case, I suppose each dapp only has 1 layout but yes there could have use case that check list layout user had installed

apps/web/scripts/kernel/handlers/kernel.ts
FME849: yes, this method could be refactored in the future if use case increased

apps/web/src/bridge/listeners/index.ts
FME849: the former will be updated with transaction decode feature, this will take more time to research

packages/sdk/providers/common.ts
FME849: these timeout maybe varying among different requests, I thought about an enum that define the timeout

apps/web/scripts/kernel/utils/popup.ts
```diff
@@ -1,11 +1,14 @@
 import type { PopupType } from '@walless/messaging';
 
-export const openPopup = async (popupType: PopupType, requestId: string) => {
+export const openPopup = async (
+	popupType: PopupType | string,
```
fluctlight-kayaba, line 4: don't use `PopupType | string`, use PopupType

apps/web/src/bridge/listeners/index.ts
```diff
@@ -93,6 +93,8 @@ export const getMessageOrTransaction = async (requestId: string) => {
 			};
 		} else if ('transaction' in res) {
 			return res;
```
fluctlight-kayaba, line 95: should comment with `// TODO: ...` to explains why

apps/web/src/screens/Request/Connection/index.tsx
```diff
@@ -25,6 +30,16 @@ const RequestConnection = () => {
...
 	}, []);
 
+	useEffect(() => {
+		getDataFromSourceRequest(requestId as string)
+			.then((result) => {
+				if (result?.sender) {
+					setSender(result.sender);
+				}
+			})
+			.catch((error) => console.log(error));
```
fluctlight-kayaba, line 40: don't use `.then`, `.catch` - use async instead.
```typescript
```

apps/web/src/screens/Request/Connection/index.tsx
```diff
@@ -35,7 +50,7 @@ const RequestConnection = () => {
 						Connection request
 					</Text>
 					<Image
-						src={logoUri}
+						src={sender?.tab?.favIconUrl || logoUri}
```
fluctlight-kayaba, line 53: `sender.tab?.favIconUrl` instead, sender is `{}` by default already.

apps/web/src/screens/Request/Connection/index.tsx
```diff
@@ -44,9 +59,9 @@ const RequestConnection = () => {
 						marginVertical={10}
 					/>
 					<Text fontSize={18} fontWeight="400">
-						Under Realm
+						{sender?.tab?.title || 'unknown'}
```
fluctlight-kayaba, line 62: sender never null

apps/web/src/screens/Request/Connection/index.tsx
```diff
@@ -44,9 +59,9 @@ const RequestConnection = () => {
...
 					</Text>
-					<LightText fontSize={14}>underrealm.stormgate.io</LightText>
+					<LightText fontSize={14}>{sender?.tab?.url || 'unknown'}</LightText>
```
fluctlight-kayaba, line 64: sender never null

apps/web/src/screens/Request/Layout/index.tsx
```diff
@@ -31,6 +36,16 @@ export const RequestLayout = () => {
...
 	}, []);
 
+	useEffect(() => {
+		getDataFromSourceRequest(requestId as string)
+			.then((result) => {
+				if (result?.sender) {
+					setSender(result.sender);
+				}
+			})
+			.catch((error) => console.log(error));
```
fluctlight-kayaba, line 46: don't use `.then`, `.catch`

apps/web/src/screens/Request/Layout/index.tsx
```diff
@@ -41,7 +56,7 @@ export const RequestLayout = () => {
 						Layout request
 					</Text>
 					<Image
-						src={logoUri}
+						src={sender?.tab?.favIconUrl || logoUri}
```
fluctlight-kayaba, line 59: sender never null

apps/web/src/screens/Request/Layout/index.tsx
```diff
@@ -50,15 +65,17 @@ export const RequestLayout = () => {
 						marginVertical={10}
 					/>
 					<Text fontSize={18} fontWeight="400">
-						Under Realm
+						{sender?.tab?.title || 'Unknown'}
```
fluctlight-kayaba, line 68: sender never null

apps/web/src/screens/Request/Layout/index.tsx
```diff
@@ -50,15 +65,17 @@ export const RequestLayout = () => {
...
-					<LightText fontSize={14}>underrealm.stormgate.io</LightText>
+					<LightText fontSize={14}>{sender?.tab?.url || 'Unknown'}</LightText>
 				</Stack>
 
 				<Stack paddingTop={30} alignItems="center">
 					<Text textAlign="center" fontSize={14} fontWeight="300">
-						Under Realm would like to add its custom layout appearance to your
-						Walless account.
+						{`${
+							sender?.tab?.title || 'Unknown'
```
fluctlight-kayaba, line 76: sender never null

apps/web/src/screens/Request/Signature/Approval.tsx
```diff
@@ -28,7 +31,7 @@ export const RequestSignatureApproval: FC<Props> = ({
 						Your signature has been requested
 					</Text>
 					<Image
-						src={logoUri}
+						src={sender?.tab?.favIconUrl || logoUri}
```
fluctlight-kayaba, line 34: sender never null

apps/web/src/screens/Request/Signature/Approval.tsx
```diff
@@ -37,9 +40,9 @@ export const RequestSignatureApproval: FC<Props> = ({
 						marginVertical={10}
 					/>
 					<Text fontSize={18} fontWeight="400">
-						Under Realm
+						{sender?.tab?.title || 'Unknown'}
```
fluctlight-kayaba, line 43: sender never null

apps/web/src/screens/Request/Signature/Approval.tsx
```diff
@@ -8,12 +9,14 @@ import LightText from 'components/LightText';
 import { logoSize, logoUri } from '../shared';
 
 interface Props {
+	sender?: UnknownObject;
```
fluctlight-kayaba, line 12: sender is required, not optional

apps/web/src/screens/Request/Signature/index.tsx
```diff
@@ -76,8 +79,9 @@ export const RequestSignature = () => {
 	}, []);
 
 	useEffect(() => {
-		getMessageOrTransaction(requestId).then((result) => {
+		getDataFromSourceRequest(requestId).then((result) => {
 			setPayload(result as MessagePayload);
+			setSender(result?.sender);
```
fluctlight-kayaba, line 84: refactor all similar places in to a re-usable Effect, e.g `useSourceRequestData`

packages/sdk/providers/common.ts
```diff
@@ -1,26 +1,50 @@
-import type { ConnectOptions } from '@walless/core';
+import { type ConnectOptions, Timeout } from '@walless/core';
```
fluctlight-kayaba, line 1: should be ```typescript
import type { type ConnectOptions } from '@walless/core';
import type { Timeout } from '@walless/core';
```

fluctlight-kayaba (approved):

PR #251, opened 2023-07-21
wl-200 solana realtime account

dev-tronghieu:


FME849 (approved):

apps/web/src/utils/hooks/wallet.ts
```diff
@@ -60,19 +60,28 @@ export const useTokens = (
...
+						ele._id.includes(address || '') &&
+						ele.count > 0,
 				);
 		}, [collectionsMap, network, address]),
 
 		collectibles: useMemo(() => {
+			const collectibles = Array.from(collectiblesMap.values()).filter(
+				(ele) => ele.account.amount > 0,
+			);
+
```
fluctlight-kayaba, line 84: prefer filter from outside of the return statement, like before:
```typescript
const collectibles = Array.from(collectiblesMap.values()).filter(...);
const collections = Array.from(collectionsMap.values()).filter(...);
```

fluctlight-kayaba (approved):

PR #256, opened 2023-07-31
feat: wl-203 under realm layout

FME849:


apps/web/src/screens/Dashboard/Embed.tsx
```diff
@@ -18,13 +19,28 @@ export const Embed: FC = () => {
...
 };
 
 export default Embed;
+
+const styles = StyleSheet.create({
+	image: {
+		width: 352,
+		height: 600,
+	},
+});
```
tanlethanh, line 46: can you make it full screen

tanlethanh (approved):

apps/web/src/screens/Dashboard/Embed.tsx
FME849: this is a mock image of Under Realm layout which contain in Dashboard view, therefore it can not be full-screen

fluctlight-kayaba (approved):

PR #265, opened 2023-08-10
chore: fix style and hide Developer section, waiting for content update

FME849:


apps/landing/pages/index.tsx
```diff
@@ -1,5 +1,5 @@
 import HomeLayout from 'components/layouts/Home';
-import Developer from 'features/home/Developer';
+// import Developer from 'features/home/Developer';
```
ltminhthu, line 2: why did you not remove it instead of commenting?

apps/landing/pages/index.tsx
```diff
@@ -15,7 +15,7 @@ export const IndexPage = () => {
 			<ExtensionDownload />
 			<EasyOnboarding />
 			<EditTool />
-			<Developer />
+			{/* <Developer /> */}
```
ltminhthu, line 18: same as above

apps/landing/pages/index.tsx
FME849: I only hide temporarily, because it has to wait for content update

ltminhthu (approved):

PR #278, opened 2023-08-30
sdk bug

tanlethanh:


apps/web/scripts/kernel/messaging.ts
```diff
@@ -38,23 +39,37 @@ export const initializeMessaging = async (): Promise<void> => {
 		const callbackRegistry: Record<string, MessengerCallback> = {};
 
 		runtime.onConnect.addListener((port) => {
-			const handleInComingMessage = async (message: EncryptedMessage) => {
+			console.log('Connecting with:', port.sender?.url);
```
fluctlight-kayaba, line 42: remove this log

apps/web/scripts/kernel/messaging.ts
```diff
@@ -38,23 +39,37 @@ export const initializeMessaging = async (): Promise<void> => {
...
+			const handleInComingMessage = async (
+				message: EncryptedMessage | MessagePayload,
+			) => {
+				console.log('On message');
```
fluctlight-kayaba, line 46: remove log

apps/web/scripts/kernel/messaging.ts
```diff
@@ -38,23 +39,37 @@ export const initializeMessaging = async (): Promise<void> => {
...
 				if (registeredCallback) {
 					if (isEncrypted) {
 						const key = await modules.encryptionKeyVault.get(port.name);
-						const decrypted = await decryptMessage(message, key);
-
+						const decrypted = await decryptMessage(
+							message as EncryptedMessage,
+							key,
+						);
+						console.log(
```
fluctlight-kayaba, line 57: remove or make it display on development only

apps/web/scripts/kernel/messaging.ts
```diff
@@ -38,23 +39,37 @@ export const initializeMessaging = async (): Promise<void> => {
...
+							`\tMessage info: \n\t - from: ${decrypted.from}, \n\t - type: ${decrypted.type} \n\t - requestId: ${decrypted.requestId}`,
+						);
 						registeredCallback?.(decrypted, port);
 					} else {
+						const payload = message as MessagePayload;
+						console.log(
```
fluctlight-kayaba, line 63: remove log

packages/core/utils/entityTypes.ts
```diff
@@ -82,6 +82,6 @@ export interface Collectible {
 
 export interface TrustedDomain {
 	trusted: boolean;
-	timestamp: string;
 	connectCount: number;
+	connect: boolean;
```
fluctlight-kayaba, line 86: does this mean `connected`?

tanlethanh: yeah will rename it

apps/web/scripts/kernel/messaging.ts
tanlethanh: Might we put this log here? Each time debug with sdk or service worker, I always need to know what happens. So if it has description, I want to put it here for later

fluctlight-kayaba: we may need an util that logs but conditional, e.g only on development - would be okay if we leave it this way for now.

fluctlight-kayaba (approved):

PR #293, opened 2023-09-04
wl-276 migrate engine module for mobile

fluctlight-kayaba:


apps/mobile/src/screens/Authentication/CreatePasscode.tsx
```diff
@@ -1,11 +1,10 @@
 import { CreatePasscode } from '@walless/app';
+import { appActions } from 'state/app';
 import { navigate } from 'utils/navigation';
 
-import { initLocalDeviceByPasscodeAndSync } from '../state/app/authentication';
-
 export const CreatePasscodeScreen = () => {
 	const handleOnComplete = async (passcode: string) => {
-		await initLocalDeviceByPasscodeAndSync(passcode);
+		await appActions.signInWithPasscode(passcode);
```
tanlethanh, line 7: meaning of this method is setup keys, encryption (in a local device) after google sign in phase. It might cause confuse to use 'signIn' again

apps/mobile/src/screens/Authentication/DeprecatedPasscode.tsx
```diff
@@ -3,15 +3,14 @@ import {
...
 import { navigate } from 'utils/navigation';
 
-import { initLocalDeviceByPasscodeAndSync } from '../state/app/authentication';
-
 export const DeprecatedPasscodeScreen = () => {
 	const handleOnSuccess = async (passcode: string) => {
 		const registeredAccount = await initAndRegisterWallet();
 		if (registeredAccount?.identifier) {
-			await initLocalDeviceByPasscodeAndSync(passcode);
+			await appActions.signInWithPasscode(passcode);
```
tanlethanh, line 13: It's like above

apps/mobile/src/screens/Splash.tsx
```diff
@@ -1,10 +1,21 @@
...
+type Props = StackScreenProps<RootParamList, 'Splash'>;
+
+export const SplashWrapper: FC<Props> = (props) => {
+	// props.navigation.navigate('Dashboard', {
+	// 	screen: 'Home',
+	// 	params: {
+	// 		screen: 'Widget',
+	// 		params: { id: 'solana' },
+	// 	},
+	// });
```
tanlethanh, line 17: do we need this code?

apps/mobile/src/utils/firebase.ts
```diff
@@ -27,16 +27,16 @@ export const fireCache: FireCache = {
...
 	if (auth().currentUser) {
 		fireCache.idToken = await auth().currentUser?.getIdToken();
+
+		auth().onIdTokenChanged(async (user) => {
+			if (user) {
+				fireCache.idToken = await user.getIdToken();
+			} else {
+				fireCache.idToken = undefined;
+			}
+		});
```
tanlethanh, line 40: I think putting event listener in a function scope might be not good. Prefer to put it at global/module scope

apps/web/src/state/app/authentication.ts
```diff
@@ -44,116 +37,50 @@ export const signInWithGoogle = async (invitationCode?: string) => {
...
+		await setProfile(profile);
 	}
 
-	// await initByPrivateKeyModule(passcode);
+	await key.reconstructKey();
 	await initBySeedPhraseModule(passcode);
 	await key.syncLocalMetadataTransitions();
-
 	modules.engine.start();
 };
```
tanlethanh, line 86: this method could be in `@walless/auth`

apps/web/src/utils/firebase.ts
```diff
@@ -47,10 +47,16 @@ export const fireCache: FireCache = {
...
+	if (auth.currentUser) {
+		fireCache.idToken = await auth.currentUser.getIdToken();
+
+		auth.onIdTokenChanged(async (user) => {
+			if (user) {
+				fireCache.idToken = await user.getIdToken();
+			} else {
+				fireCache.idToken = undefined;
+			}
+		});
```
tanlethanh, line 60: the same comment with `mobile` above

apps/web/src/utils/ioc.ts
```diff
@@ -3,17 +3,23 @@ import { modules } from '@walless/ioc';
 import { createEncryptionKeyVault } from '@walless/messaging';
 import { configure, create } from '@walless/store';
 import IDBPouch from 'pouchdb-adapter-idb';
+import { initializeAuth } from 'utils/firebase';
+
+import { makeConfig } from '../../scripts/kernel/utils/config';
```
tanlethanh, line 8: Is `makeConfig` need to be put into common utils?

packages/auth/utils/passcode.ts
```diff
@@ -19,3 +19,9 @@ export const validateAndRecoverWithPasscode = async (passcode: string) => {
 	const unlockSuccess = await recoverDeviceShareFromPasscode(passcode);
 	return unlockSuccess;
 };
+
+/**
+ * @deprecated
+ * This method is deprecated by using recovery flow
```
tanlethanh, line 25: `method` to `constant`

packages/auth/utils/recovery.ts
```diff
@@ -1,17 +1,18 @@
...
 
-import { injectedModules } from './ioc';
+import { key } from './w3a';
 
 export const initAndRegisterWallet = async (): Promise<Account | undefined> => {
 	try {
-		await injectedModules.key.reconstructKey();
-		const newShare = await injectedModules.key.generateNewShare();
+		await key().reconstructKey();
+		const newShare = await key().generateNewShare();
```
tanlethanh, line 11: why do we need key as a method?

packages/crypto/utils/passcode.ts
tanlethanh, line 1: I see `passcode` for web below, why dont we name it for `mobile` only?

apps/mobile/src/screens/Authentication/DeprecatedPasscode.tsx
fluctlight-kayaba: lets discuss this in above comment

apps/mobile/src/utils/firebase.ts
fluctlight-kayaba: hmm, agree -- you reminds me that we'll miss the case that use login after open the App, above code block won't covered that case, while always subscribe the change won't take that much effort/resource at global/module scope anyway.

apps/web/src/utils/ioc.ts
fluctlight-kayaba: yes, because it is used in more than 1 single place

packages/auth/utils/recovery.ts
fluctlight-kayaba: its a way to bind aliased type for `never`.

apps/mobile/src/screens/Authentication/CreatePasscode.tsx
tanlethanh: using `signInWith...` looks like sign in by different methods than stages. Otherwise, with deprecated passcode, action to use passcode is actually 'sign in' by look up `tkey question module`. But using 'passcode' to setup local encryption might be not sign in (still a stage).

PR #298, opened 2023-09-26
update: wl-483 sdk connection is unstable

FME849:


FME849:
Will need to refactor Request UI to free extension from tamagui

apps/web/scripts/background/index.ts
```diff
@@ -1,3 +1,26 @@
...
+		}
+
+		// Add a noop listener to the alarm. Without this, the service worker seems
+		// to be deemed as idle by Chrome and will be killed after 30s.
+		chrome.alarms.onAlarm.addListener(() => {
+			Function.prototype();
+		});
+	}
+})();
+
```
fluctlight-kayaba, line 23: don't put the method here, make it a method somewhere and import to this `index.ts` file

fluctlight-kayaba (approved):

apps/web/scripts/background/index.ts
FME849: After discussion, we decided that there was no better place to put this method, so just a small refactoring of the code to make it easier to read

fluctlight-kayaba (approved):

PR #299, opened 2023-10-05
feat: wl-452 universal sui assets

FME849:


packages/engine/crawlers/sui/index.ts
```diff
@@ -18,33 +17,15 @@ export const suiEngineRunner: SuiRunner = {
...
-				const coin = allCoins[i];
-				const object = allObjects[i];
-
-				docPromises[i] = getTokenDocument(
-					context,
-					owner,
-					coin,
-					object?.data as never,
-				);
+				docPromises[i] = getTokenDocument(context, owner, coin);
```
fluctlight-kayaba, line 28: make sure that we can avoid multiple request, if that is possible then should send as less request as possible

fluctlight-kayaba (approved):

PR #303, opened 2023-10-23
feat: wl-452 universal sui assets

FME849:


packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+import type { CollectibleDocument } from '@walless/store';
+
+export const getSuiCollectibles = async (
+	connection: JsonRpcProvider,
+	owner: string,
+) => {
+	const collectibleDocuments: CollectibleDocument[] = [];
+	let objectResponses: PaginatedObjectsResponse['data'] = [];
+	let nextCursor: PaginatedObjectsResponse['nextCursor'] | undefined;
+	let cursor: PaginatedObjectsResponse['nextCursor'] | undefined;
```
tanlethanh, line 16: Do we need to mark it undefine when declare uninitialized variables

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+
+		cursor = allOwnedObjects.nextCursor;
+
+		if (allOwnedObjects.hasNextPage && nextCursor !== cursor) {
+			nextCursor = cursor;
+		} else {
+			nextCursor = null;
+		}
+
+		objectResponses = [...objectResponses, ...allOwnedObjects.data];
```
tanlethanh, line 37: In this part of code, the main logic is `objectResponses = [...objectResponses, ...allOwnedObjects.data];` should be placed on top of remaining which is checking next cursor.

Also, does it really need `cursor` variable, I think just assign nextCusor continuously, if it does not have next, it will be null or undefined

```
		cursor = allOwnedObjects.nextCursor;

		if (allOwnedObjects.hasNextPage && nextCursor !== cursor) {
			nextCursor = cursor;
		} else {
			nextCursor = null;
		}
``` 
to
```
nextCursor = allOwnedObjects.nextCursor;
```

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	objectResponses
+		.filter((objResponse) => {
+			if (objResponse.data) {
+				return isCollectible(objResponse.data);
+			}
+
+			return false;
+		})
```
tanlethanh, line 47: isCollectible should return false if arg is null/undefined. So it just `objectResponsed.filter(res => isCollectible(res.data))`

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+		.forEach((objResponse) => {
+			if (objResponse.data) {
+				collectibleDocuments.push(
```
tanlethanh, line 50: After filter, dont need to recheck here

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+	return collectibleDocuments;
+};
+
+const getCollectibleDocument = (
+	objResponse: SuiObjectData,
+	owner: string,
+): CollectibleDocument => {
+	const collectibleId = `${owner}/${objResponse.objectId}`;
+	const formatMetadata = () => {
+		const metadata: CollectibleDocument['metadata'] = {
```
tanlethanh, line 65: We already have Metadata type exported in @walless/core I guess

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+			sod: objResponse,
+		};
+		if (
+			objResponse.display?.data &&
+			typeof objResponse.display.data === 'object'
+		) {
+			const objectDisplay = objResponse.display.data;
+			if ('name' in objectDisplay) {
+				metadata.name = objectDisplay.name;
+			}
```
tanlethanh, line 78: It can be `if (objectDisplay.name) { ...}` or `metadata.name = objectDisplay.name || "Unknown"` and dont init at above (just empty object when init metadata)

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+				metadata.description = objectDisplay.description;
+			}
+		}
+
+		return metadata;
+	};
+
+	return {
+		_id: collectibleId,
+		collectionId: 'Unknown',
```
tanlethanh, line 94: Add a to do comment here to easily find it later `// TODO: check nft collection on SUI`

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+
+	return collectibleDocuments;
+};
+
+const getCollectibleDocument = (
+	objResponse: SuiObjectData,
+	owner: string,
+): CollectibleDocument => {
+	const collectibleId = `${owner}/${objResponse.objectId}`;
+	const formatMetadata = () => {
```
tanlethanh, line 64: Why do we need this function declaration, I saw that it was not re-used anywhere

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+		return true;
+	}
+
+	let url: string | undefined;
+	if (
+		suiObject.display?.data &&
+		typeof suiObject.display.data === 'object' &&
+		('image_url' in suiObject.display.data ||
+			'img_url' in suiObject.display.data ||
+			'url' in suiObject.display.data)
```
tanlethanh, line 117: This type of condition is not familiar I think ~~, just like above comments

packages/engine/crawlers/sui/collectibles.ts
```diff
@@ -0,0 +1,151 @@
...
+
+	return false;
+};
+
+const isCollection = (suiObject: SuiObjectData): boolean => {
+	return (
+		!!suiObject.content &&
+		'fields' in suiObject.content &&
+		'logical_owner' in suiObject.content.fields &&
+		'bag' in suiObject.content.fields
```
tanlethanh, line 149: same comment as above

packages/engine/crawlers/sui/subscription.ts
```diff
@@ -22,6 +25,19 @@ export const suiTokenSubscribe = (
...
+	connection: JsonRpcProvider,
+	owner: string,
+) => {
+	suiSubscriptionPool.push(
+		setInterval(async () => {
+			const collectibleDocs = await getSuiCollectibles(connection, owner);
+
+			collectibleActions.setCollectibles(collectibleDocs);
+		}, 30000),
+	);
```
tanlethanh, line 38: mark to todo comment to optimize it later I think

packages/engine/crawlers/sui/collectibles.ts
FME849: `cursor` is needed to be undefined for being used in the first `while` loop, avoid the `used before being assigned` error

packages/engine/crawlers/sui/collectibles.ts
FME849: As `suiObject.content` could be variable types which could be not include field `fields` therefore checking if `fields`  in object is necessary, other fields can be check normally

packages/engine/crawlers/sui/subscription.ts
FME849: Sui has not support onChange events, already checked with Sui Foundation and explored Ethos, Suiet wallets, so this is the possible solution right now

tanlethanh (approved):

PR #308, opened 2023-10-27
update: wl-525 persistent assets

FME849:


packages/engine/crawlers/solana/index.ts
```diff
@@ -56,10 +59,14 @@ export const solanaEngineRunner: SolanaRunner = {
...
-			tokenActions.setItems(fungibleTokens);
+			tokenActions.setTokens(fungibleTokens);
+			solanaCollectiblesByAddress({
+				context,
+				address: currentPubkey.toString(),
+			});
 
 			for (const key of accountKeys) {
-				registerAccountChanges(context, liveConnection, key);
+				registerAccountChanges(context, liveConnection, key, currentPubkey);
```
tanlethanh, line 69: with passing the context, consider to remove `liveConnection`, this connection changes the commitment for request, we could reconstruct the new connection each time we need (it might not take performance effect)

packages/engine/crawlers/solana/collectibles.ts
```diff
@@ -70,11 +74,13 @@ export const addCollectibleToState = async (
 		collectionMetadata?.address.toString() ||
 		nft.collection?.address.toString() ||
 		nft.address.toString();
-	const collectionId = `${address}/${collectionAddress}`;
+	const collectionId = `${address}/collection/${collectionAddress}`;
+	const collectibleId = `${address}/collectible/${nft.mint.address.toString()}`;
 
 	const collection = collectionState.map.get(collectionId);
+	const collectible = collectibleState.map.get(collectibleId);
```
tanlethanh, line 81: for now, I think we just need to use the storage, decouple from state. That means we will not use state at engine level anymore. It will make logical flow cleaner, just update storage. Any storage change will affect to app state by storage listener.
this comment also means we need to refactor this engine by just using storage

tanlethanh: @fluctlight-kayaba pls have a look

packages/engine/crawlers/solana/collectibles.ts
```diff
@@ -70,11 +74,13 @@ export const addCollectibleToState = async (
...
 	if (!collection) {
-		collectionState.map.set(collectionId, {
+		collectibleActions.setStorageCollection(collectionId, {
```
tanlethanh, line 83: same as above comment, and for all the actions change state

packages/engine/crawlers/solana/collectibles.ts
```diff
@@ -90,12 +96,14 @@ export const addCollectibleToState = async (
...
+	} else if (!collectible) {
+		collectibleActions.updateCollectionAmount(
+			collectionId,
+			(collection.count += 1),
+		);
 	}
 
-	const collectibleId = `${address}/${nft.mint.address.toString()}`;
-	collectibleState.map.set(collectibleId, {
+	collectibleActions.setStorageCollectible(collectibleId, {
```
tanlethanh, line 106: I think storage is in data level, not in the app state level, so just make it a single function instead of use as a state action

packages/engine/crawlers/solana/index.ts
```diff
@@ -43,7 +44,9 @@ export const solanaEngineRunner: SolanaRunner = {
 				const tokenType = getTokenType(account);
 
 				if (tokenType === TokenType.Fungible) {
-					fungiblePromises.push(getSPLTokenDocument(context, pubkey, account));
+					fungiblePromises.push(
```
tanlethanh, line 47: Should we rename it to `tokenPromises`? (different from collections and collectibles)

packages/engine/crawlers/solana/index.ts
```diff
@@ -56,10 +59,14 @@ export const solanaEngineRunner: SolanaRunner = {
 				item.account.quotes = quotes[makeHashId(item)].quotes;
 			}
 
-			tokenActions.setItems(fungibleTokens);
+			tokenActions.setTokens(fungibleTokens);
+			solanaCollectiblesByAddress({
```
tanlethanh, line 63: use `Promise.all` to wait for collectibles setup?

packages/engine/crawlers/solana/subscription.ts
```diff
@@ -40,18 +42,29 @@ export const registerAccountChanges = async (
 			balance = data.amount.toString();
 		}
 
-		const id = `${address}/${mint}`;
+		const tokenId = `${owner}/token/${mint}`;
+		const isToken = tokenState.map.get(tokenId);
```
tanlethanh, line 46: not use state anymore?

packages/engine/crawlers/solana/subscription.ts
```diff
@@ -40,18 +42,29 @@ export const registerAccountChanges = async (
...
 
-		if (!tokenActions.updateBalance(id, balance)) {
+		if (isToken) {
+			tokenActions.updateBalance(tokenId, balance);
```
tanlethanh, line 49: not use state anymore?

packages/engine/crawlers/solana/subscription.ts
```diff
@@ -40,18 +42,29 @@ export const registerAccountChanges = async (
...
-			if (!collectibleActions.updateCollectibleAmount(id, amount)) {
+			const collectibleId = `${owner}/collectible/${mint}`;
+
+			if (amount === 0) {
+				collectibleActions.removeCollectibleDoc(collectibleId);
+			} else if (
+				!(await collectibleActions.updateCollectibleAmount(
+					collectibleId,
+					amount,
+				))
```
tanlethanh, line 60: using storage updating instead?

packages/engine/crawlers/solana/subscription.ts
```diff
@@ -87,16 +100,18 @@ export const watchLogs = async (
...
 			);
 
 			for (const balance of tokenBalances || []) {
 				if (balance.uiTokenAmount.decimals === 0) {
-					const mpl = new Metaplex(connection);
-					const nft = await mpl
-						.nfts()
-						.findByMint({ mintAddress: new PublicKey(balance.mint) });
-					addCollectibleToState(connection, endpoint, address, nft);
+					if (balance.uiTokenAmount.amount === '1') {
```
tanlethanh, line 108: you can use `if (balance.uiTokenAmount.decimals === 0 && balance.uiTokenAmount.amount === '1')`

tanlethanh: also consider check it !== '0', for semi-fungible assets?

packages/engine/crawlers/solana/subscription.ts
```diff
@@ -136,7 +151,7 @@ export const watchLogs = async (
 						metadata,
 					};
 
-					tokenState.map.set(tokenDocument._id, tokenDocument);
+					tokenActions.setTokens([tokenDocument]);
```
tanlethanh, line 154: not use state anymore?

packages/engine/crawlers/sui/index.ts
```diff
@@ -41,7 +41,7 @@ export const suiEngineRunner: SuiRunner = {
 				item.account.quotes = quotes[makeHashId(item)].quotes;
 			}
 
-			tokenActions.setItems(tokenDocs);
+			tokenActions.setTokens(tokenDocs);
```
tanlethanh, line 44: not use state anymore?

packages/engine/crawlers/tezos/index.ts
```diff
@@ -28,7 +28,7 @@ export const tezosEngineRunner: TezosRunner = {
 			item.account.quotes = quotes[makeHashId(item)].quotes;
 		}
 
-		tokenActions.setItems(tokenDocs);
+		tokenActions.setTokens(tokenDocs);
```
tanlethanh, line 31: not use state anymore?

packages/engine/state/collectible/index.ts
```diff
@@ -13,20 +14,61 @@ export const collectibleActions = {
...
+		);
+		const collectionCount = collection!.count - 1;
+
+		if (collectionCount === 0) {
+			await modules.storage.removeDoc(collection!._id);
+		} else {
+			await modules.storage.upsert<CollectionDocument>(
+				collection!._id,
+				async (prevDoc) => {
+					prevDoc.count -= 1;
```
tanlethanh, line 64: these functions are in data storage level, so I prefer putting them into storage package

packages/engine/state/token/index.ts
```diff
@@ -8,13 +9,26 @@ export const tokenActions = {
...
+		const result = await modules.storage.upsert<TokenDocument>(
+			id,
+			async (prevDoc) => {
+				prevDoc.account.balance = balance;
+
+				return prevDoc;
+			},
+		);
+
+		return result.ok;
```
tanlethanh, line 31: same as the previous comment

packages/engine/crawlers/solana/index.ts
tanlethanh: I found the problem of `liveConnection`, It seems like we missunderstood about commitment, for listen account change faster on solana, we use confirm commitment. But instead of reconstruct the connection, we could add optional argument `commitment` when listen to account change. So pls refactor this one by removing live connection, and change any liveConnection using to add `confirm` in the listening function call.

packages/engine/crawlers/solana/index.ts
FME849: collectibles and collections is non-fungible token, so I guess it's fine. However, I suggest `splPromises`

packages/engine/crawlers/solana/index.ts
FME849: already refactored and used `Promise.all` in `setTokens` method

packages/engine/crawlers/solana/token.ts
```diff
@@ -55,3 +65,51 @@ export const getSPLTokenDocument = async (
...
+				getSPLTokenDocument(context, pubkey, account, ownerPubkey),
+			);
+		}
+
+		accountKeys.push(pubkey);
+	}
+
+	const fungibleTokens = (await Promise.all(fungiblePromises)).filter(
+		(token) => token.account.mint === solMint || token.account.balance !== '0',
+	);
```
FME849, line 103: this filter out the tokens which balance = 0 in `fungibleTokens`, but in other chain like tezos, we continue to display tokens with 0 balance, do we need to unify this UI issue? @tanlethanh

packages/engine/crawlers/solana/collectibles.ts
```diff
@@ -12,23 +12,29 @@ import { PublicKey } from '@solana/web3.js';
...
 import { Networks } from '@walless/core';
 
-import { collectibleState, collectionState } from '../../state/collectible';
+import {
+	getCollectibleById,
+	getCollectionById,
+	setCollectible,
+	setCollection,
+	updateCollectionAmount,
+} from '../../utils/collectibles';
```
tanlethanh, line 21: I prefer putting these functions in 'storage' packages

packages/engine/crawlers/solana/index.ts
```diff
@@ -1,66 +1,33 @@
...
+import { solanaFungiblesByAddress } from './token';
 
 export const solanaEngineRunner: SolanaRunner = {
 	start: async (context) => {
 		const { storage } = modules;
-		const { connection } = context;
-		const liveConnection = new Connection(connection.rpcEndpoint, 'confirmed');
-		const key = await storage.find<PublicKeyDocument>(selectors.solanaKeys);
+		const key = await storage.find<PublicKeyDocument>(
+			selectors.solanaKeys as never,
```
tanlethanh, line 16: why do we need to use `as never` here?

packages/engine/crawlers/tezos/index.ts
```diff
@@ -2,20 +2,23 @@ import { modules } from '@walless/ioc';
...
 import { KNOWN_TEZOS_MAINNET_TOKENS } from './shared';
-import { getTokenDocument } from './token';
+import { getNativeTokenDocument, getTokenDocument } from './token';
 
 export const tezosEngineRunner: TezosRunner = {
 	start: async (context) => {
 		const { storage } = modules;
-		const key = await storage.find<PublicKeyDocument>(selectors.tezosKeys);
+		const key = await storage.find<PublicKeyDocument>(
+			selectors.tezosKeys as never,
```
tanlethanh, line 16: I see a lot of using `as never`, I think it should be resolve by another way

packages/engine/crawlers/tezos/index.ts
```diff
@@ -28,7 +31,7 @@ export const tezosEngineRunner: TezosRunner = {
 			item.account.quotes = quotes[makeHashId(item)].quotes;
 		}
 
-		tokenActions.setItems(tokenDocs);
+		setTokens(tokenDocs);
```
tanlethanh, line 34: I am confuse with this `setTokens`, does it update state or storage. If about storage, prefer to putting it at storage package

packages/engine/crawlers/tezos/shared.ts
```diff
@@ -30,23 +30,7 @@ export const getTezosEndpointFromUnifiedEndpoint = (endpoint: Endpoints) => {
...
-	},
-	account: {
-		address: 'tezos-native-token',
-		balance: '0',
-		decimals: 6,
-	},
-};
-
-export const KNOWN_TEZOS_MAINNET_TOKENS: Partial<TokenDocument>[] = [
+export const KNOWN_TEZOS_MAINNET_TOKENS: Omit<Token, 'network'>[] = [
```
tanlethanh, line 33: like it :v

packages/engine/state/live.ts
```diff
@@ -10,21 +12,34 @@ import type {
...
 	const { storage } = modules;
-	const widget = await storage.find<WidgetDocument>(f.allWidgets);
-	const allKey = await storage.find<PublicKeyDocument>(f.allKeys);
-	const allToken = await storage.find<TokenDocument>(f.allTokens);
+	const widget = await storage.find<WidgetDocument>(f.allWidgets as never);
+	const allKey = await storage.find<PublicKeyDocument>(f.allKeys as never);
+	const allToken = await storage.find<TokenDocument>(f.allTokens as never);
+	const allCollectibles = await storage.find<CollectibleDocument>(
+		f.allCollectibles as never,
+	);
```
tanlethanh, line 31: need to resolve types of selectors

packages/engine/utils/collectibles.ts
tanlethanh, line 1: I think this utils should be put in `storage`

packages/engine/utils/token.ts
tanlethanh, line 1: same as above comment

packages/engine/utils/token.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const setTokens = (tokens: TokenDocument[]) => {
+	const tokenPromises: Promise<unknown>[] = [];
+
+	for (const token of tokens) {
+		tokenPromises.push(modules.storage.upsert(token._id, async () => token));
+	}
+
+	Promise.all(tokenPromises);
+};
```
tanlethanh, line 16: I prefer name it `create` or `insert`, use `setTokens` makes me think it's a set state action

packages/engine/utils/collectibles.ts
```diff
@@ -0,0 +1,91 @@
...
+	return await modules.storage.safeGet(id);
+};
+
+const setCollectible = async (id: string, item: CollectibleDocument) => {
+	return await modules.storage.upsert(id, async () => item);
+};
+
+const setCollection = async (id: string, item: CollectionDocument) => {
+	return await modules.storage.upsert(id, async () => item);
+};
```
tanlethanh, line 22: I prefer name it create or insert, use setTokens makes me think it's a set state action

packages/store/utils/helper.ts
tanlethanh, line 1: need to resolve types

packages/engine/crawlers/tezos/index.ts
FME849: `setTokens` working with storage not state

packages/engine/utils/token.ts
FME849: could use `addTokensToStorage`, I guess?

packages/store/utils/helper.ts
FME849: do you have any ideas, these types help reduce typo error when creating selectors, but it's the root cause of those `as never` type

packages/engine/utils/collectibles.ts
FME849: can you explain why, is it because these methods are related to storage? but I think they're engine methods and not re-used some where else

packages/engine/state/live.ts
FME849: I think the problem is `find` method because of it's type `<never | undefined>`, whatever type you give to selectors, it'll always show errors. What do you think?

packages/engine/utils/token.ts
tanlethanh: if using this name, we could rename all of functions make change storage make adding suffix `ToStorage1

packages/store/utils/helper.ts
tanlethanh: I'll check it

packages/engine/utils/collectibles.ts
tanlethanh: Yes, It's totally related to storage.  It can be used without knowing what is engine, might be use anywhere it need in the future

packages/store/utils/helper.ts
FME849: I'm waiting for solution to complete this PR =]]] @tanlethanh

tanlethanh: It was resolved, by just removing the SelectorFunc which we are not using, and also make types un-manageable

PR #315, opened 2023-11-06
integrate gasilon to mobile

ltminhthu:


apps/mobile/src/utils/transaction.ts
```diff
@@ -30,10 +36,19 @@ export const createAndSend: CreateAndSendFunction = async (
...
+				res.signatureString = await solanaHandlers.signAndSendTransaction(
+					transaction,
+					privateKey,
+				);
+			} else {
+				res.signatureString =
+					await solanaHandlers.signTransactionAbstractionFee(
+						transaction,
+						privateKey,
+					);
```
tanlethanh, line 49: does it sign and also send the transaction?

apps/web/scripts/kernel/handlers/solanaHandler.ts
```diff
@@ -95,15 +95,18 @@ export const signTransactionAbstractionFee: HandleMethod = async ({
...
+				},
+				body: JSON.stringify({
+					transaction: txStr,
+				}),
 			},
-			body: JSON.stringify({
-				transaction: txStr,
-			}),
-		});
+		);
```
tanlethanh, line 109: is it reusable for both web and mobile?

tanlethanh: if we have this function in package app, prefer removing it

packages/kernel/handlers/solanaHandler.ts
```diff
@@ -60,3 +60,36 @@ export const signAndSendTransaction = async (
 
 	return signatureStr;
 };
+
+export const signTransactionAbstractionFee = async (
```
tanlethanh, line 64: `signAndSendAbstractionFeeTransaction` I think

apps/mobile/src/utils/transaction.ts
ltminhthu: yes, should i change the name?

apps/web/scripts/kernel/handlers/solanaHandler.ts
ltminhthu: > if we have this function in package app, prefer removing it

you said that you need to refactor request message on web platform so that i cannot remove this file immediately

PR #327, opened 2023-11-13
temporary fix ResizeObserver does not exist in mobile

tanlethanh:


packages/gui/components/ModalManager/ModalContainer.tsx
```diff
@@ -86,20 +91,22 @@ export const ModalContainer: FC<Props> = ({ item }) => {
...
+					bindingRectangle: updatedBindingRectangle,
+				});
 			});
-		});
-		bindingObserver.observe(actualBindingRef.current as never);
+			bindingObserver.observe(actualBindingRef.current as never);
 
-		return () => bindingObserver.disconnect();
+			return () => bindingObserver.disconnect();
+		}
```
fluctlight-kayaba, line 109:
```typescript 
if (Platform.OS !== 'web') return;
...keep other code as it is
```
then we could revert easier, less line changes

fluctlight-kayaba (approved):

PR #338, opened 2023-11-18
feat: wl-449 backbone for analytics

fluctlight-kayaba:


dev-tronghieu (approved):

apps/web/src/utils/firebase.ts
```diff
@@ -47,16 +53,24 @@ export const fireCache: FireCache = {
...
+				fireCache.idToken = await user.getIdToken();
+
+				if (appState.remoteConfig.deepAnalyticsEnabled) {
+					await universalActions.syncRemoteProfile();
+				}
+			} else {
+				fireCache.idToken = undefined;
+			}
+		});
+	}
```
tanlethanh, line 73: this auth event listener should not be initialized in `initializeAuth`, when users sign in, they need to reload the app to sync the analytics. Need to put it in global scope or module scope.

tanlethanh (approved):

PR #359, opened 2023-11-22
hotfix: just register wallet after completing passcode at init status

tanlethanh:


packages/auth/utils/signIn.ts
```diff
@@ -39,13 +38,14 @@ export const signInWithTorusKey = async ({
...
-		if (registeredAccount?.identifier) {
-			await handlePasscode();
-		} else {
-			await handleError();
-		}
+		// if (registeredAccount?.identifier) {
+		// 	await handlePasscode();
+		// } else {
+		// 	await handleError();
+		// }
```
fluctlight-kayaba, line 48: delete this if no-longer needed

PR #371, opened 2023-12-01
[WL-785] Fix bug passcode

ltminhthu:


packages/app/features/Passcode/PasscodeInput/SingleInput.tsx
```diff
@@ -58,6 +58,7 @@ const SingleInput = forwardRef<TextInput, Props>(
 					onKeyPress={onKeyPress}
 					onFocus={() => setIsFocused(true)}
 					onBlur={() => setIsFocused(false)}
+					textContentType="creditCardNumber"
```
FME849, line 61: this props only work on iOS, should use something universal

PR #380, opened 2023-12-06
[WL-690] Update connect popup UI

dev-tronghieu:


apps/web/src/screens/Request/Connect.tsx
```diff
@@ -85,13 +85,17 @@ const RequestConnect = () => {
 					<LightText textAlign="center">
 						Only connect to websites you trust!
 					</LightText>
-					<Button title="Connect" onPress={onApprovePress}></Button>
+					<Button
+						hitSlop={{ top: 20, bottom: 20, left: 20, right: 20 }}
```
fluctlight-kayaba, line 89: just `hitSlop={20} `

fluctlight-kayaba (approved):

apps/web/src/screens/Request/Connect.tsx
dev-tronghieu: Here is the `Button` props
```tsx
interface Props {
	style?: StyleProp<ViewStyle>;
	children?: ReactNode;
	title?: string;
	titleStyle?: StyleProp<TextStyle>;
	onPress?: () => void;
	outline?: boolean;
	transparent?: boolean;
	disabled?: boolean;
	hitSlop: Insets;  // [Fluctlight Kayaba], 6 days ago   (November 30th, 2023 7:45 PM)
}
```

1. `hitSlop` is required, do you want to change it to optional?
Screenshot 2023-12-06 at 16 41 24

2. Your `hitSlop` prop is declared as `Insets` type, do you want to change it to `number | Insets`?
Screenshot 2023-12-06 at 16 41 44

---

The updated `Button` props can be
```tsx
interface Props {
	style?: StyleProp<ViewStyle>;
	children?: ReactNode;
	title?: string;
	titleStyle?: StyleProp<TextStyle>;
	onPress?: () => void;
	outline?: boolean;
	transparent?: boolean;
	disabled?: boolean;
	hitSlop?: Insets | number;
}
```

fluctlight-kayaba: Please also update `hitSlop` props here, I think we want to remove it.
**Note:** even Button did not declare a default `hitSlop` value.. it's underlying component `Hoverable` had, so even if we leave `hitSlop` it will fallback to a reasonable value already.

FME849 (approved):

PR #389, opened 2023-12-10
feat: pwa build optimization

fluctlight-kayaba:


apps/web/package.json
```diff
@@ -52,6 +52,7 @@
     "raf": "^3.4.1",
     "react-native-qrcode-svg": "^6.2.0",
     "react-native-reanimated": "*",
+    "react-native-safe-area-context": "4.7.4",
```
FME849, line 55: can we move this dependency to root `node_modules` and use in `peerDependencies` as it's used by both in packages and apps?

apps/web/src/screens/Explore.tsx
```diff
@@ -1,8 +1,88 @@
-import type { FC } from 'react';
-import { ExplorerFeature } from '@walless/app';
+import { type FC, useMemo, useState } from 'react';
```
FME849, line 1: fix this import

yarn.lock
FME849, line 1: why yarm lock change a lot?

tanlethanh (approved):

PR #396, opened 2023-12-14
fix: [WL-848] select component animation

FME849:


packages/gui/states/modal/type.ts
```diff
@@ -58,6 +58,7 @@ export interface ModalConfigs {
 	withoutMask?: boolean;
 	hide?: boolean;
 	context?: unknown;
+	updater?: () => void;
```
fluctlight-kayaba, line 61: something like `onBeforeHide` would explain this better.

PR #405, opened 2024-01-03
feat: experiment moving packages/app to apps/mobile

FME849:


packages/gui/index.ts
```diff
@@ -7,6 +7,7 @@ export * from './components/DrawerContainer';
 export * from './components/Hoverable';
 export * from './components/Input';
 export * from './components/ModalManager';
+export * from './components/Passcode';
```
dev-tronghieu, line 10: I don't think we want to put `Passcode` in `@walless/gui`, this package is only used for general UI components, right? `Passcode` seems like only belonged to `apps/wallet`.

Need a confirmation from @cloudle

tanlethanh: I think we can put `passcode` here, just a single component with its logic, without any understanding about feature

apps/mobile/src/modals/Error.tsx
```diff
@@ -0,0 +1,29 @@
...
+		justifyContent: 'center',
+		alignItems: 'center',
+		backgroundColor: '#AE3939',
+		paddingTop: Math.max(insets.top, 20),
+		paddingBottom: 20,
+	};
+
+	return (
+		<View style={containerStyle}>
+			<Text>{(config as UnknownObject).context.errorText}</Text>
```
tanlethanh, line 24: why do we need the `as UnknownObject` here?, if without error message, prefer to set "Something went wrong"

tanlethanh (approved):

PR #412, opened 2024-01-08
refactor: clean up Receive modal & add Explorer feature

dev-tronghieu:


apps/mobile/src/modals/Receive/index.tsx
```diff
@@ -17,12 +17,9 @@ import { useSnapshot } from 'valtio';
...
 import { modalStyles } from '../internal';
 
-import Slider, {
-	type IndicatorOption,
-	type SlideOption,
-} from './components/Slider';
-import WalletCard, { type WalletProps } from './components/WalletCard';
-import WalletCardIndicator from './components/WalletCardIndicator';
+import Indicator from './Indicator';
+import Item, { type WalletProps } from './Item';
```
tanlethanh, line 21: I prefer using another name for a good understanding of it

tanlethanh (approved):

apps/mobile/src/modals/Receive/index.tsx
dev-tronghieu: Can you suggest some names? These two components are internal components of the `Receive` modal, they are not exposed anywhere else, hence I think simple names such as `Item` and `Indicator` are just fit with the internal `Slider` in the same scope.

PR #413, opened 2024-01-08
invitation feature and log in feature, fix small details

ltminhthu:


tanlethanh (commented):
for Invitation, and Login, I prefer putting it into `screens` instead of `features`. `features` is used for reusable feature with a lot of logic.

FME849 (approved):

PR #418, opened 2024-01-09
complete remove @walless/app from mobile - universal app

tanlethanh:


apps/mobile/src/screens/Splash.tsx
```diff
@@ -38,9 +38,11 @@ export const SplashScreen: FC = () => {
...
 		};
 
-		Promise.all([appActions.bootstrap(), playAnimate()]).then(() => {
-			appActions.launchApp();
-		});
+		Promise.all([appActions.bootstrap?.(), playAnimate()]).then(
+			([response]) => {
+				appActions.launchApp?.(response);
+			},
+		);
```
FME849: `appActions.bootstrap` is not undefined, so removed `?.`

FME849 (approved):

PR #420, opened 2024-01-09
refactor: ioc config

dev-tronghieu:


apps/mobile/src/utils/ioc/index.ts
```diff
@@ -32,7 +31,6 @@ export const injectModules = async () => {
 
 	modules.native = nativeModules;
 	modules.analytics = universalAnalytics;
-	modules.config = Config;
```
FME849, line 35: Is`environment` in `utils/config` able to use in native platform?

FME849 (approved):

PR #432, opened 2024-01-22
[WL-884]refactor: sending nft flow and package network

FME849:


packages/network/src/solana/transactions.ts
```diff
@@ -52,3 +79,131 @@ export const simulateTransaction = async (
...
+	const message = new TransactionMessage({
+		payerKey: new PublicKey(sender),
+		recentBlockhash: blockhash,
+		instructions,
+	}).compileToV0Message();
+
+	return new VersionedTransaction(message);
+};
+
+export const constructSendSPLTokenTransactionInSol = async (
```
tanlethanh, line 109: I prefer to separate network for different network, like solana. We can import by `@walless/network/solana` or using `export * as solana` in the index file of network package, then we will have `import { solana } from `@walless/network` and use like `solana.constructSendSPLTokenTransaction` without `inSol`

packages/network/src/solana/transactions.ts
```diff
@@ -52,3 +79,131 @@ export const simulateTransaction = async (
...
+	connection: Connection,
+	sender: PublicKey,
+	receiver: PublicKey,
+	amount: number,
+	token: Token | Collectible,
+) => {
+	// ATA
+	const mintAddress = new PublicKey(token.account.mint as string);
```
tanlethanh, line 117: from now, mint is required in token account, so you can remove as string here

packages/network/src/solana/transactions.ts
FME849: InSol mean to differentiate token to pay fee between native Sol and SPL token through gasilon

packages/network/src/solana/transactions.ts
```diff
@@ -52,3 +79,131 @@ export const simulateTransaction = async (
...
+		instructions.push(transferInstruction);
+	} else {
+		const transferInstruction = createTransferTokenInstruction(
+			senderATAddress,
+			receiverATAddress,
+			sender,
+			amount,
+		);
+		instructions.push(transferInstruction);
+	}
```
tanlethanh, line 196: I think we should keep the previous function and implement another one for the new standard, the checking standard needs to be in the upper scope which calls these functions.

packages/network/src/solana/transactions.ts
```diff
@@ -1,9 +1,36 @@
...
-import { Keypair, Transaction, VersionedTransaction } from '@solana/web3.js';
+import { PublicKey, SYSVAR_INSTRUCTIONS_PUBKEY } from '@solana/web3.js';
+import {
+	Keypair,
+	SystemProgram,
+	Transaction,
+	TransactionMessage,
+	VersionedTransaction,
+} from '@solana/web3.js';
+import type { Collectible, Token } from '@walless/core';
```
tanlethanh, line 22: I think the network package should not know about other stuff of @walless, it's just a wrapper for utils when working with network like Solana

tanlethanh (approved):

PR #440, opened 2024-01-24
[WL-844] Fullscreen Modal & Keyboard Avoiding View

devtronghieu:


apps/wallet/src/components/KeyboardAvoidingView.tsx
```diff
@@ -0,0 +1,37 @@
...
+				{children}
+			</TouchableWithoutFeedback>
+		</BaseKeyboardAvoidingView>
+	);
+};
+
+const styles = StyleSheet.create({
+	container: {
+		display: 'flex',
+		flexDirection: 'column',
```
tanlethanh, line 32: I think it's a default style of a View component

packages/gui/components/ModalManager/ModalContainer.tsx
```diff
@@ -73,6 +75,11 @@ export const ModalContainer: FC<Props> = ({ item }) => {
 			opacity: opacity.value,
 		};
 		if (fullWidth) baseStyle.width = width.value;
+		if (fullHeight) {
+			baseStyle.height = height.value;
+			baseStyle.top = 0;
+			baseStyle.left = 0;
```
tanlethanh: we should remove `left` because it does not relate to height binding. Also, we currently treat modal as bottom sheet, so binding the bottom to 0 is more suitable than the top.

packages/gui/components/SwipeDownGesture.tsx
```diff
@@ -17,7 +21,7 @@ import { AnimatedView } from './aliased';
 
 interface Props {
 	children: ReactNode;
-	style?: ViewStyle;
+	style?: ViewStyle | ViewStyle[];
```
tanlethanh, line 21: In the use of this style, it's already used in an array with animatedStyle, you need to check if it's possible to use array of styles for this prop

tanlethanh (approved):

PR #442, opened 2024-01-24
[WL-844] Fix web input cannot type

devtronghieu:


apps/wallet/src/components/KeyboardAvoidingView.tsx
```diff
@@ -14,6 +15,10 @@ interface Props {
 }
 
 const KeyboardAvoidingView: FC<Props> = ({ children, style }) => {
+	if (Platform.OS === 'web') {
+		return <View style={[styles.container, style]}>{children}</View>;
+	}
+
```
tanlethanh, line 21: try to use .web file to replace this one. Should platform separation at build time instead of runtime

PR #445, opened 2024-01-24
[WL-967] Bug - navigate after send nft

ltminhthu:


apps/wallet/src/features/Send/InputTransaction/index.tsx
```diff
@@ -1,10 +1,11 @@
-import type { FC } from 'react';
+import { type FC, useEffect } from 'react';
```
tanlethanh, line 1: user `import type {}` instead

apps/wallet/src/features/Send/InputTransaction/index.tsx
```diff
@@ -30,6 +33,16 @@ const InputTransaction: FC<Props> = ({ navigator }) => {
...
 
+	useEffect(() => {
+		if (collectible) {
+			txActions.update({ type: 'Collectible' });
+			const collection = collections.find(
+				(ele) => ele._id === collectible.collectionId,
+			);
+			txActions.update({ collection });
+		}
+	}, []);
```
tanlethanh, line 44: should include collectible to the dependencies, each time the collectible change -> auto update correspond collection

apps/wallet/src/features/Send/TransactionResult/index.tsx
```diff
@@ -22,6 +23,23 @@ const TransactionResult: FC<Props> = ({ navigator }) => {
...
+		navigationRef.reset({
+			index: 1,
+			routes: [
+				{ name: 'Dashboard' },
+				{
+					name: 'Dashboard',
+					params: {
+						screen: 'Home',
+					},
+				},
```
tanlethanh, line 36: I don't know what we call `home`, but if I send an NFT from Solana widget, I prefer to back to the Solana screen instead of to the dashboard. You can get this by querying the latest path, which is also the latest widget

apps/wallet/src/features/Send/TransactionResult/index.tsx
```diff
@@ -22,6 +23,23 @@ const TransactionResult: FC<Props> = ({ navigator }) => {
...
+			],
+		});
+		txActions.resetTransactionContext();
+		txActions.closeSendFeature();
```
tanlethanh, line 40: `closeSendFeature` has already called `resetTransactionContext`

apps/wallet/src/screens/Dashboard/Collectible.tsx
```diff
@@ -37,7 +52,7 @@ export const CollectibleFeat = () => {
 	useEffect(() => {
 		// TODO: need to resolve and remove can go back if possible
 		if (!curCollectible && navigationRef.canGoBack()) {
-			// navigateBack();
+			handleBackToHome();
```
tanlethanh, line 55: please make sure that this one is not called twice, the problem will come when you sent an NFT and try to click on the `collection`. Need to recheck this one, it's a separated bug

apps/wallet/src/utils/hooks/navigation.ts
```diff
@@ -21,10 +21,21 @@ export const useNavigationHydrate = () => {
...
+					params: {
+						id: string;
+					};
+				};
+			};
+
+			const widgetId =
+				childRoute?.params.screen === 'Collectible' || 'Collection'
+					? ''
+					: childRoute.params?.params?.id;
```
FME849: your solution will always redirect users to `Explore` screen, that's unexpected behavior

apps/wallet/src/utils/hooks/navigation.ts
```diff
@@ -21,10 +21,21 @@ export const useNavigationHydrate = () => {
...
 
 			if (childRoute?.name === 'Explore' && widgetId !== lastWidgetId) {
```
FME849, line 41: I find that there is a simpler way, just keep `widget` variable, and add an additional condition here `widget.screen === 'Widget'` because what we want is ignore all routes except widget route.

apps/wallet/src/features/Send/InputTransaction/index.tsx
```diff
@@ -30,6 +34,16 @@ const InputTransaction: FC<Props> = ({ navigator }) => {
...
 
+	useEffect(() => {
+		if (collectible) {
+			txActions.update({ type: 'Collectible' });
+			const collection = collections.find(
+				(ele) => ele._id === collectible.collectionId,
+			);
+			txActions.update({ collection });
+		}
+	}, [collectible]);
```
FME849, line 45: I suggest using method `getCollectionByIdFromStorage` in `utils/storage/collectible.ts` to make code more transparent

apps/wallet/src/features/Send/TransactionResult/index.tsx
FME849: idk why this marked as resolved because there is no change from last preview, and I suggest trying `resetRoute` method in `utils/navigation/index.ts`

apps/wallet/src/utils/hooks/navigation.ts
devtronghieu: You mean this?

```ts
const widgetId = childRoute?.params?.screen === 'Widget'
	? childRoute.params?.params.id
	: '';
```

apps/wallet/src/features/Send/TransactionResult/index.tsx
devtronghieu: This function is used for the Button "Back to home", so its mission is to nav back to the `Home` screen.

`resetRoute` is a better approach, I have tweaked the code to use that method

apps/wallet/src/features/Send/InputTransaction/index.tsx
devtronghieu: please check the latest change

apps/wallet/src/screens/Dashboard/CollectionStack/index.tsx
```diff
@@ -0,0 +1,46 @@
+import type { FC } from 'react';
```
tanlethanh, line 1: We had a `stacks` dir for defining navigation system, could define these stacks inside the `stacks` dir, and keep this screen's structure

ltminhthu: > We had a `stacks` dir for defining navigation system, could define these stacks inside the `stacks` dir, and keep this screen's structure
okay, i'll move it to `stacks/Explorer`

apps/wallet/src/features/Widget/BuiltInNetwork/CollectiblesTab.tsx
```diff
@@ -18,9 +18,16 @@ export const CollectiblesTab: FC<Props> = ({ collections = [] }) => {
...
 	const handlePressItem = (ele: WrappedCollection) => {
 		const collectionId = ele._id.split('/')[2];
+
 		navigate('Dashboard', {
 			screen: 'Explore',
-			params: { screen: 'Collection', params: { id: collectionId } },
+			params: {
+				screen: 'Collection',
+				params: {
+					screen: 'Default',
```
tanlethanh, line 27: can you explain the `Default` screen

apps/wallet/src/stacks/Explorer/CollectionStack/Collectible.tsx
```diff
@@ -1,24 +1,34 @@
...
 	}, [collectibles]);
 
-	const curCollection = useMemo(() => {
-		return collections.find((ele) => ele._id === curCollectible?.collectionId);
+	useMemo(() => {
+		if (!curCollectible) return;
+
+		setCurCollection(
+			collections.find((ele) =>
+				ele._id.includes(curCollectible?.collectionId || 'invalid-collection'),
```
tanlethanh, line 29: do not need 'invalid-collection', you return the callback after checking null for `curCollectible`. Also, remove the question mark

apps/wallet/src/stacks/Explorer/CollectionStack/Collection.tsx
```diff
@@ -23,13 +22,29 @@ export const CollectionFeat: FC = () => {
...
+				screen: 'Collection',
+				params: {
+					screen: 'Collectible',
+					params: { id },
+				},
+			},
 		});
 	};
 
+	useEffect(() => {
```
tanlethanh, line 42: `useFocusEffect` here might be better

apps/wallet/src/stacks/Explorer/CollectionStack/Collectible.tsx
```diff
@@ -29,9 +39,8 @@ export const CollectibleFeat = () => {
 	};
 
 	useEffect(() => {
-		// TODO: need to resolve and remove can go back if possible
-		if (!curCollectible && navigationRef.canGoBack()) {
-			// navigateBack();
+		if (!curCollectible) {
+			navigateBack();
```
tanlethanh, line 43: `useFocusEffect` with binding navigation effect

apps/wallet/src/stacks/Explorer/index.tsx
```diff
@@ -26,16 +24,6 @@ export const ExplorerStack = () => {
...
-	const ManagedCollectionScreen = withStackContainer(CollectionScreen, {
-		title: 'Collection',
-		goBack: navigateBack,
-	});
-
-	const ManagedCollectibleScreen = withStackContainer(CollectibleScreen, {
-		title: 'Collectible',
-		goBack: navigateBack,
-	});
-
```
tanlethanh, line 38: I see things here, the problem might be caused by `withStackContainer`, please checking it again

apps/wallet/src/stacks/Explorer/index.tsx
```diff
@@ -45,12 +33,8 @@ export const ExplorerStack = () => {
...
 			<Drawer.Screen
 				name="Collection"
-				component={ManagedCollectionScreen}
-				options={options}
-			/>
-			<Drawer.Screen
-				name="Collectible"
-				component={ManagedCollectibleScreen}
+				component={CollectionStack}
+				initialParams={{ screen: 'Default' }}
```
tanlethanh, line 37: CollectionStack for Collectible might cause confusing

apps/wallet/src/utils/hooks/navigation.ts
```diff
@@ -21,10 +21,21 @@ export const useNavigationHydrate = () => {
 			/* eslint-disable */
 			const findDashboard = (i: any) => i.name === 'Dashboard';
 			const dashboard = state?.routes.find(findDashboard)?.state;
-			const childRoute = dashboard?.routes[dashboard?.index as never];
-			const widget = childRoute?.params as any;
-			const widgetId = widget?.params?.id;
-			/* eslint-enable */
```
tanlethanh, line 27: keep eslint-enable or remove eslint-diable if not using `any` here

tanlethanh (changes requested):

tanlethanh:
@cloudle you should review this PR

apps/wallet/src/features/Widget/BuiltInNetwork/CollectiblesTab.tsx
ltminhthu: i think we'll have a collection stacks which contains the collection screen (default screen) and collectible screen, so that is the way i treat it a default screen

apps/wallet/src/stacks/Explorer/index.tsx
ltminhthu: what problem are you concerned about?

apps/wallet/src/utils/hooks/navigation.ts
ltminhthu: @devtronghieu is that yours?

apps/wallet/src/features/Send/InputTransaction/index.tsx
```diff
@@ -30,6 +34,16 @@ const InputTransaction: FC<Props> = ({ navigator }) => {
 		}
 	};
 
+	useEffect(() => {
+		if (collectible) {
+			txActions.update({ type: 'Collectible' });
+			const collection = collections.find(
+				(ele) => ele._id === collectible.collectionId,
+			);
+			txActions.update({ collection });
```
fluctlight-kayaba, line 43: 1. seems that `txActions.update` would be call just once, instead of 2 times like this
2. if the whole idea if this effect is to keep `txContext` (a valtio instance) and something managed by `txActions` the logic shouldn't be here as an effect under component level - we could e.g subscribe for changes from valtio and sync it with others in `bootstrap` or engine's setup.

apps/wallet/src/features/Send/TransactionResult/index.tsx
```diff
@@ -22,6 +23,22 @@ const TransactionResult: FC<Props> = ({ navigator }) => {
...
+			routes: [
+				{ name: 'Dashboard' },
+				{
+					name: 'Dashboard',
+					params: {
+						screen: 'Home',
+					},
+				},
+			],
+		});
```
fluctlight-kayaba, line 38: this should be handled/implement inside utils/navigation, then other places could re-use the same method.

apps/wallet/src/stacks/Dashboard/TabBar.tsx
```diff
@@ -45,16 +47,18 @@ export const BottomNavigationTabBar: FC<Props> = ({ tabProps }) => {
...
-		const { routes, index } = state;
-		/* eslint-disable-next-line */
-		const currentRoute: any = routes[index];
-		const widgetId = currentRoute.params?.params?.id;
-		const isExploreTab = currentRoute.name === 'Explore' && !widgetId;
+		const mustHaveBottomTabBar = mustHaveBottomTabBarRoutes.includes(
+			currentRoute.name,
+		);
+
+		const nextOffset = isDrawerOpen || mustHaveBottomTabBar ? 0 : realBarHeight;
```
fluctlight-kayaba, line 58: would this cover `Widget Explorer` screen? this doesn't seem related or part of this Ticket anyway.

apps/wallet/src/stacks/Explorer/CollectionStack/Collectible.tsx
```diff
@@ -1,24 +1,34 @@
...
 	const {
 		params: { id = '' },
 	} = useRoute() as never;
 
 	const curCollectible = useMemo(() => {
 		if (!id) return;
-		return collectibles.find((ele) => ele._id.includes(id));
+
+		const collectible = collectibles.find((ele) => ele._id.includes(id));
+		return collectible;
```
fluctlight-kayaba, line 21: please keep the old code here.

apps/wallet/src/stacks/Explorer/CollectionStack/Collectible.tsx
```diff
@@ -1,24 +1,34 @@
...
 	}, [collectibles]);
 
-	const curCollection = useMemo(() => {
-		return collections.find((ele) => ele._id === curCollectible?.collectionId);
+	useMemo(() => {
```
fluctlight-kayaba, line 24: useMemo being called/used as it is useEffect, use it this way cause confusion understanding the code.

apps/wallet/src/stacks/Explorer/CollectionStack/Collectible.tsx
```diff
@@ -29,9 +39,8 @@ export const CollectibleFeat = () => {
 	};
 
 	useEffect(() => {
-		// TODO: need to resolve and remove can go back if possible
-		if (!curCollectible && navigationRef.canGoBack()) {
-			// navigateBack();
+		if (!curCollectible) {
```
fluctlight-kayaba, line 42: use this kind of hook to check for e.g when to navigation usually a bad idea, it would become extremely complex quickly.

apps/wallet/src/stacks/Explorer/index.tsx
ltminhthu: i've already removed it

apps/wallet/src/features/Send/InputTransaction/index.tsx
ltminhthu: > 1. seems that `txActions.update` would be call just once, instead of 2 times like this
> 2. if the whole idea if this effect is to keep `txContext` (a valtio instance) and something managed by `txActions` the logic shouldn't be here as an effect under component level - we could e.g subscribe for changes from valtio and sync it with others in `bootstrap` or engine's setup.

1. I'll fix it
2. It would be component level because there are some specific cases that we need to check whether it's collectible tx (currently there is a case in collectible screen)

apps/wallet/src/features/Send/TransactionResult/index.tsx
ltminhthu: I think the `reset` method of react native navigation is generic enough to use instead of rewriting it in `utils/navigation`

apps/wallet/src/stacks/Dashboard/TabBar.tsx
ltminhthu: > would this cover `Widget Explorer` screen? this doesn't seem related or part of this Ticket anyway, not even related.

because it has an impact on the way i change the navigation of collection and collectible inside a widget so that i have to adjust it
the `Widget Explorer` is not included in the `TabBar` so I think we don't need to cover it

tanlethanh (approved):

tanlethanh (approved):

PR #448, opened 2024-01-29
[WL-658, WL-965] feat: persistent Solana history 

FME849:


apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+import type { ParsedTokenAccountWithAddress } from './types';
+
+const historyLimit = 20;
+
+export const getTransactionsHistory = async (
+	connection: Connection,
+	wallet: PublicKey,
+	accounts: ParsedTokenAccountWithAddress[],
+) => {
+	const splTokenAccount = accounts
```
tanlethanh, line 27: `splTokenAccounts`

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+	});
+};
+
+const getTransactionSignatures = async (
+	connection: Connection,
+	accountList: PublicKey[],
+) => {
+	const signatureBufferList = await Promise.all(
+		accountList.map(async (account) => {
+			const transactionList = await connection.getSignaturesForAddress(
```
tanlethanh, line 67: for this query, I think we need to add pthrottle, It will throttle network requests to prevent quota limitation of Solana RPC

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+export const getTransactionsHistory = async (
+	connection: Connection,
+	wallet: PublicKey,
+	accounts: ParsedTokenAccountWithAddress[],
+) => {
+	const splTokenAccount = accounts
+		.filter((account) => account.tokenAmount.decimals > 0)
+		.map((account) => account.publicKey);
+
+	const accountList = [wallet].concat(splTokenAccount);
```
tanlethanh, line 31: It's good to use built-in array method, but It looks weird because `concat` is usually used to merge two arrays into a new array (we can also do this by `[...a, ...b]`. In this case, you constructed a new array to merge, so the good one for this is using `push`. `splTokenAccounts.push(wallet)`, and if pushing the wallet account makes `splTokenAccounts` become non-sense, you should rename it just `tokenAccounts` I guess

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+	const signatureList = [
+		...new Set(await getTransactionSignatures(connection, accountList)),
+	];
+	const transactions = (
+		await getTransactionsDetail(connection, signatureList)
+	).slice(0, 20);
```
tanlethanh, line 37: you should slice the signatures instead of fetching details first to slice, It will cause a performance problem

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+		)
+	).filter((doc) => doc !== null);
+
+	removeOldHistories(transactionHistoryDocuments);
+
+	transactionHistoryDocuments.forEach((historyDoc) => {
+		storage.upsert<TransactionHistoryDocument>(historyDoc._id, async () => {
+			return historyDoc;
+		});
+	});
```
tanlethanh, line 58: insert new documents first and then remove the old ones later, If user is in History view before completing engine start phase, It might cause unexpected behavior

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+				},
+			);
+
+			return transactionList.map((trans) => trans.signature);
+		}),
+	);
+
+	const signatureList = signatureBufferList.reduce((signatureList, subList) => {
+		return signatureList.concat(subList);
+	}, []);
```
tanlethanh, line 80: you are flattening a 2d array? If using flatten, I think js should have built-in method to do that, or using lodash if possible

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+			maxSupportedTransactionVersion: 0,
+			commitment: 'finalized',
+		},
+	);
+	const sortedTransactions = parsedTransactions.sort((trans1, trans2) =>
+		(trans1?.blockTime || 0) < (trans2?.blockTime || 0) ? 1 : -1,
+	);
+
+	return sortedTransactions;
+};
```
tanlethanh, line 101: I saw this function does two things, get parsed txs and sort them by time, I don't think it suits the name, `getTransactionsDetailsInBlockTimeOrder` might be suitable, or you might not need this function, just put it into the calling one because it's not reusable in another place

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+		token,
+	};
+};
+
+const analyzeTokenBalances = (
+	preTokenBalances: TokenBalance[],
+	postTokenBalances: TokenBalance[],
+) => {
+	if (postTokenBalances.length > preTokenBalances.length) {
+		const referralPreBalance = preTokenBalances[0];
```
tanlethanh, line 220: Why does it include `referral` here

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+		sender,
+		receiver,
+	};
+};
+
+const getTransactionFee = async (walletAddress: string) => {
+	return await storage.safeGet<TokenDocument>(
+		`${walletAddress}/token/${solana.solMint}`,
+	);
+};
```
tanlethanh, line 284: why is it named `getTransactionFee`, I saw it just gets the native token document

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+) => {
+	const { docs } = await storage.find(selectors.allHistories);
+
+	if (docs.length === 0) return;
+
+	const oldHistoryDocs = (docs as TransactionHistoryDocument[]).sort(
+		(doc1, doc2) => (doc1.date > doc2.date ? -1 : 1),
+	);
+
+	for (let i = 0; i < 20; i++) {
```
tanlethanh, line 297: should not have magic number here

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+		} else {
+			oldHistoryDocs.shift();
+		}
+	}
+
+	Promise.all(
+		oldHistoryDocs.map(async (doc) => {
+			await storage.removeDoc(doc._id);
+		}),
+	);
```
tanlethanh, line 317: if using `Promise.all`, It should be aware that it needs to use await to have correct behavior, If a function calls and waits for this one, it will not wait for complete removing

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -0,0 +1,318 @@
...
+
+		if (currentNewDoc) {
+			if (currentNewDoc._id === currentOldDoc._id) {
+				oldHistoryDocs.shift();
+			}
+			newHistoryDocs.shift();
+		} else {
+			oldHistoryDocs.shift();
+		}
+	}
```
tanlethanh, line 311: Does it remove the oldest first?

apps/wallet/src/engine/runners/solana/history.ts
FME849: this method means to get the transaction's tokenForFee, in the future, we might fetch other token using in Gasilon transactions. I renamed it to `getTransactionTokenFee` for more transparent meaning

apps/wallet/src/engine/runners/solana/history.ts
FME849: this is my mistake, I mean referenced. Because `preTokenBalances` and `postTokenBalanced` are not always identical, I need to choose 1 `TokenBalance` as a referenced to perform calculation. Update to `referencedPreBalance` etc.

apps/wallet/src/engine/runners/solana/history.ts
FME849: Because `signatureList` is signatures from several token accounts that is not in order, it is inconvenient to slice the signatures

apps/wallet/src/engine/runners/solana/history.ts
FME849: this function mean to remove old history that out of `historyLimit` range, but I already refactor this function

tanlethanh (approved):

PR #449, opened 2024-01-29
[WL-910] arc refactor, continued

fluctlight-kayaba:
- Messaging module for Mobile (WIP)
- Bare minimal Browser tab for Mobile


apps/wallet/browser/kernel/handlers/common.ts
```diff
@@ -1,16 +1,16 @@
-import type { ConnectOptions } from '@walless/core';
 import { Networks } from '@walless/core';
-import { ResponseCode } from '@walless/messaging';
+import { ResponseCode } from '@walless/core';
+import type { ConnectOptions } from '@walless/sdk';
 import type { PublicKeyDocument, TrustedDomainDocument } from '@walless/store';
 import { selectors } from '@walless/store';
+import { storage } from 'utils/storage/db';
```
tanlethanh, line 6: I was wondering if we import `utils/storage`in `browser`. For kernel only, I created a storage instance in `apps/wallet/browser/kernel/utils/storage.ts`. I think we should keep platforms or environments separately

apps/wallet/browser/bridge/utils.ts
```diff
@@ -2,7 +2,7 @@ import { logger } from '@walless/core';
 import type { PureMessagePayload } from '@walless/messaging';
 import { createEncryptionKeyVault, createMessenger } from '@walless/messaging';
 import type { SettingDocument } from '@walless/store';
-import { storage } from 'utils/storage';
+import { storage } from 'utils/storage/db';
```
tanlethanh, line 5: This bridge is used for web app communication, can we use a separate implementation to decouple code from another place?

apps/wallet/browser/kernel/index.ts
```diff
@@ -1,15 +1,14 @@
 import { logger, runtime } from '@walless/core';
+import { configure } from '@walless/store';
+import { storage } from 'utils/storage/db';
```
tanlethanh, line 3: as same as the above comment

apps/wallet/browser/kernel/utils/helper.ts
```diff
@@ -1,14 +1,16 @@
...
 import { decryptWithPasscode } from '@walless/crypto';
-import { RequestType } from '@walless/messaging';
-import type { ExtensionDocument } from '@walless/store';
-import type { PrivateKeyDocument, PublicKeyDocument } from '@walless/store';
+import type {
+	ExtensionDocument,
+	PrivateKeyDocument,
+	PublicKeyDocument,
+} from '@walless/store';
+import { storage } from 'utils/storage/db';
```
tanlethanh, line 8: as same as the above comment

apps/wallet/browser/kernel/utils/middleware.ts
```diff
@@ -1,17 +1,18 @@
...
 	RequestType,
 	ResponseCode,
 	ResponseMessage,
-} from '@walless/messaging';
+} from '@walless/core';
 import { utils } from '@walless/network';
+import type { ConnectOptions } from '@walless/sdk';
 import type { TrustedDomainDocument } from '@walless/store';
 import { selectors } from '@walless/store';
+import { storage } from 'utils/storage/db';
```
tanlethanh, line 12: as same as the above comment

apps/wallet/index.js
```diff
@@ -1,8 +1,10 @@
-import { AppRegistry } from 'react-native';
+require('utils/shim');
```
tanlethanh, line 1: can you explain why using `require` here?

packages/core/utils/platform.ts
```diff
@@ -26,10 +20,10 @@ export const runtime: UniversalRuntime = {
...
 	isExtension,
 	isServer,
-	isIOS,
-	isAndroid,
-	isMobile,
-	isBrowser: isWeb && !isExtension && !isServer,
+	isBrowser: !isExtension && !isServer,
+	isMobile: false,
+	isIOS: false,
+	isAndroid: false,
```
tanlethanh, line 26: could we consider removing this `runtime` variable when it does not reflect runtime correctly anymore

tanlethanh (approved):

PR #450, opened 2024-01-29
[WL-975] Fix "Max" button & Rework UX of input checking

devtronghieu:


tanlethanh (approved):

tanlethanh:
@devtronghieu LGTM, I approved the PR, but we need to wait for the UX content before merging this PR

apps/wallet/src/components/CheckedInput.tsx
```diff
@@ -1,31 +1,44 @@
...
-	const handlerBlur = () => {
-		setErrorText(checkFunction(value));
-	};
+	value?: string;
+	placeholder?: string;
+	keyboardType?: KeyboardTypeOptions;
+	errorText?: string;
+	suffix?: ReactNode;
+	onChangeText?: (value?: string) => void;
+	onBlur?: (value?: string) => void;
```
tanlethanh, line 13: `onBlur` should not have a param, just pass it like the original `onBlur`. It's like exposing `onBlur` prop from the component

tanlethanh: also, you should keep the props when wrapping a component if you don't override it

apps/wallet/src/components/CheckedInput.tsx
```diff
@@ -1,31 +1,44 @@
...
-				{...inputProps}
+				placeholder={placeholder}
 				value={value}
-				importantStyle={!!errorText && styles.errorInputStyle}
-				onBlur={handlerBlur}
+				inputStyle={!!errorText && styles.errorText}
+				importantStyle={!!errorText && styles.errorInputContainer}
+				keyboardType={keyboardType}
+				suffix={suffix}
+				onChangeText={onChangeText}
```
tanlethanh, line 34: you're not overriding these props, keep as previous

apps/wallet/src/features/Send/ConfirmByPasscode/index.tsx
```diff
@@ -52,6 +52,10 @@ const PasscodeInput: FC<Props> = ({ navigator, item, activatedId }) => {
 
 			try {
 				const res = await createAndSend(payload, passcode);
+				if (res.error) {
+					throw res.error;
+				}
+
```
tanlethanh, line 58: does not throw error here, createAndSend should throw Error inside it (for web/ext, it will throw error from messaging checking, double check the implementation on mobile)

packages/network/src/handlers/solana.ts
```diff
@@ -84,5 +84,9 @@ export const signAndSendTransactionAbstractionFee = async (
 
 	const data = await res.json();
 
+	if (data.status === 'error') {
+		throw new Error(data.message);
+	}
+
```
tanlethanh, line 90: should check the returned status code inside the response instead of checking body (I think it can be changed).

tanlethanh: you might want to check `res.ok`

apps/wallet/src/features/Send/ConfirmByPasscode/index.tsx
devtronghieu:
```ts
			try {
				const res = await createAndSend(payload, passcode);

				txActions.update({ time: new Date() });
				txActions.update({ status: res.responseCode });

				if (res.responseCode == ResponseCode.WRONG_PASSCODE) {
					showError({ errorText: 'Passcode is NOT matched' });
					setError('Wrong passcode');
				} else if (res.responseCode == ResponseCode.SUCCESS) {
					const signature =
						res.signatureString || res.signedTransaction?.digest || res.hash;
					txActions.update({ signatureString: signature });
					navigator.slideNext();
				} else if (res.responseCode == ResponseCode.ERROR) {
					// In the case the responseCode is Error and it does not trigger that `catch` block,
					// What should I do?
					// May I throw error here?
				}
			} catch (error) {
				showError({ errorText: (error as Error).message });
			}

			setPasscode('');
		} else if (passcode.length > 0 && error) {
			setError('');
		}
```

devtronghieu: If not handle the `res.responseCode == ResponseCode.ERROR`, the user will be stuck at the `Passcode` screen

tanlethanh: for the web, we handled throwing errors with Error Response from the messaging module, you need to throw the error on the mobile (kernel handling) to make things sync

apps/wallet/src/components/CheckedInput.tsx
```diff
@@ -1,31 +1,33 @@
...
-	const handlerBlur = () => {
-		setErrorText(checkFunction(value));
-	};
-
+const CheckedInput: FC<Props> = ({
+	value,
+	errorText,
+	onFocusOut,
+	...inputProps
+}) => {
```
tanlethanh, line 16: when I detached this component, I thought that the `errorText` was just used in this component, so keeping the state here might be good at rendering performance.

apps/wallet/src/components/CheckedInput.tsx
```diff
@@ -1,31 +1,33 @@
 import type { FC } from 'react';
-import { useState } from 'react';
 import { StyleSheet, Text, View } from 'react-native';
 import type { InputProps } from '@walless/gui';
 import { Input } from '@walless/gui';
 
 type Props = {
-	checkFunction: (value?: string) => string | undefined;
+	errorText?: string;
+	onFocusOut?: (value?: string) => void;
```
tanlethanh, line 8: Also, I think it's not good to bind value param to `onFocusOut`, as I mention below, keeping some state inside this component might be better

apps/wallet/src/features/Send/InputTransaction/CollectiblesTab.tsx
```diff
@@ -65,11 +93,21 @@ export const CollectiblesTab: FC<Props> = ({ onContinue }) => {
 				selectedItemIconStyle={styles.selectedNftIcon}
 			/>
 
-			<RecipientInput />
+			<CheckedInput
+				value={receiver}
+				placeholder="Recipient account"
+				errorText={recipientInput.message}
```
tanlethanh, line 99: Same topic with about, the `recipientInput.message` is only used for this component, try to keep things abstract, and separate.

apps/wallet/src/features/Send/InputTransaction/TokensTab.tsx
```diff
@@ -20,16 +23,23 @@ interface Props {
...
-		? parseFloat(token.account.balance) / 10 ** token.account.decimals
-		: -1;
+	const [recipientInput, setRecipientInput] = useState({
+		valid: false,
+		message: '',
+	});
+	const [amountInput, setAmountInput] = useState({
+		valid: false,
+		message: '',
+	});
```
tanlethanh, line 39: the same comment as above, prefer keeping previous verson

apps/wallet/src/features/Send/InputTransaction/TransactionFee/AbstractedTransactionFee.tsx
```diff
@@ -119,12 +116,29 @@ export const AbstractedTransactionFee: FC<Props> = ({ tokenList }) => {
...
+			const amountBN = new BN(parseFloat(amount || '0') * decimalsMultiplier);
+			isNotEnoughToken = amountBN.add(feeBN).gt(balanceBN);
+		} else {
+			isNotEnoughToken = feeBN.gt(balanceBN);
+		}
+
+		setError(
+			isNotEnoughToken
+				? `Not enough ${
+						tokenForFee.metadata?.symbol ?? 'Unknown'
```
tanlethanh, line 137: assign this to a variable to make it clean

apps/wallet/src/features/Send/InputTransaction/TransactionFee/index.tsx
```diff
@@ -13,13 +14,20 @@ interface Props {
...
+	if (isAbstractFee) {
+		const tokenList = tokens;
+		const solIndex = tokenList.findIndex(
+			(token) => token.account.mint === solMint,
+		);
+		if (solIndex !== -1) {
+			[tokenList[0], tokenList[solIndex]] = [tokenList[solIndex], tokenList[0]];
+		}
+		return <AbstractedTransactionFee tokenList={tokens} />;
+	}
```
tanlethanh, line 28: First, I think a conditional return like this is not good (does not clean).
`tokenList` should be renamed to `sortedTokens` to make it clear.
Also, I am sure that `SOL` token is always placed at the head of `tokens`. Clone a state array and mutate it to passing as a prop for child component might cause side effect

apps/wallet/src/features/Send/InputTransaction/internal.ts
```diff
@@ -1,84 +1,3 @@
 import { Networks } from '@walless/core';
-import { checkValidAddress } from 'utils/transaction';
-
-import { txContext } from '../context';
-
-export const totalCheckFieldsToContinue = () => {
```
tanlethanh, line 6: you removed the total check and disabled the button, make sure that's the correct UX as we defined or expected

apps/wallet/src/utils/transaction/common.ts
```diff
@@ -42,9 +43,17 @@ export const checkValidAddress = (keyStr: string, network: Networks) => {
...
+			message: '',
+		};
 	} catch (error) {
-		return { valid: false, message: (error as Error).message };
+		return {
+			valid: false,
+			message: network
+				? `Wrong [${capitalize(network)}] wallet address. Please check again.`
+				: (error as Error).message,
+		};
```
tanlethanh, line 56: to make things clear, I think we should refactor it to remove the valid flag, just return a string if It has error (error message)

packages/network/src/handlers/solana.ts
```diff
@@ -84,5 +84,9 @@ export const signAndSendTransactionAbstractionFee = async (
 
 	const data = await res.json();
 
+	if (!res.ok) {
+		throw new Error(data.message);
+	}
```
tanlethanh, line 89: make sure that you checked this change

apps/wallet/src/features/Send/InputTransaction/TransactionFee/index.tsx
devtronghieu: `SOL` is not placed at the head in my case (while testing, `USDC-Dev` is the first one).  I don't want to break other code so I sort it out here.

`sortedTokens` is obviously a better name, gonna change it now

tanlethanh: If `SOL` is not placed at the head, it might be a bug, but it was resolved on the newest engine updating

apps/wallet/src/utils/transaction/common.ts
devtronghieu: there is a case where `message` is `''` but it is not valid (as we update the checking condition right in the input phase), e.g. empty amount or empty recipient

tanlethanh: what next if `message === ''` but it is not valid

apps/wallet/src/features/Send/InputTransaction/CollectiblesTab.tsx
```diff
@@ -98,7 +98,7 @@ export const CollectiblesTab: FC<Props> = ({ onContinue }) => {
 				placeholder="Recipient account"
 				errorText={recipientInput.message}
 				onChangeText={(receiver) => txActions.update({ receiver })}
-				onFocusOut={handleFocusOutRecipient}
+				onBlur={() => handleFocusOutRecipient(receiver)}
```
tanlethanh, line 101: `handleFocusOutRecipient` could get the value from state, does not need to passing it. So it's just `onBlue={handleFocusOutRecipient}`

apps/wallet/src/utils/transaction/common.ts
tanlethanh: It might confuse because of using two fields for a single purpose as error checking. If the message is not empty, it still is invalid (despite the valid having true value)

tanlethanh: You should have an alter convention/pattern to make it clean

PR #451, opened 2024-01-29
[WL-910] Bug nfts not showing on mobile

devtronghieu:


apps/wallet/src/engine/runners/solana/collectibles.ts
```diff
@@ -92,21 +92,18 @@ export const loadCollectibleMetadata = async (
 		tokenOwner: owner,
 	});
 
-	// TODO: need to resolve fetch metadata of nft using metaplex on mobile
-	if (metadata.jsonLoaded && !metadata.json) {
```
devtronghieu, line 96: @FME849 I wonder if we need to check for the `metadata.jsonLoaded`? I have logged it out and its value is `false` so that there is no fetch happened

FME849: I checked and found that `metadata.jsonloaded` condition was not necessary, so removing it is fine

apps/wallet/src/engine/runners/solana/collectibles.ts
```diff
@@ -92,21 +92,18 @@ export const loadCollectibleMetadata = async (
...
-			nft = {
-				...nftByMint,
-				json: await jsonRes.json(),
-				jsonLoaded: true,
-			};
-		}
+	if (!metadata.json && 'mintAddress' in metadata) {
+		const nftByMint = await mpl.nfts().findByMint({
+			mintAddress: metadata.mintAddress,
+			loadJsonMetadata: metadata.jsonLoaded,
```
FME849, line 98: because we removed `metadata.jsonLoaded` condition, we could also remove it here.

tanlethanh (approved):

PR #452, opened 2024-01-30
[WL-981] Remove widget bug

devtronghieu:


apps/wallet/src/stacks/Explorer/WidgetNavigator/index.tsx
```diff
@@ -1,13 +1,11 @@
-import type { FC } from 'react';
+import { type FC } from 'react';
```
tanlethanh, line 1: use `type { FC }` instead

tanlethanh (approved):

PR #453, opened 2024-01-30
[WL-983] Bug mobile nfts not showing

devtronghieu:


apps/wallet/src/engine/runners/solana/collectibles.ts
```diff
@@ -159,21 +158,27 @@ export const updateRelatedCollection = async (
 		collectible.collectionId,
 	);
 
-	if (!storedCollection) {
-		const collectionMetadata: GenericNft = await mpl.nfts().findByMint({
+	if (
+		!storedCollection ||
+		!storedCollection.metadata ||
+		!('name' in storedCollection.metadata)
```
FME849, line 164: what is the reason behind this condition: field `name` not in `storedCollection.metadata`

devtronghieu: there is a case that `storedCollection` is an empty `{}` and as you know, `!{}` is `false` so it will pass the `!storedCollection.metadata` check

FME849 (approved):

PR #458, opened 2024-01-31
[WL-876] feat: implement swap on solana

tanlethanh:


apps/wallet/src/features/Swap/InputSwap/FromToken.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+import SelectButton from './SelectButton';
+
+const FromToken = () => {
+	const { fromToken, amount } = useSnapshot(swapContext).swap as SwapContext;
+
+	const balance = useMemo(() => {
+		if (!fromToken) return 0;
+		return (
+			parseFloat(fromToken.account.balance) / 10 ** fromToken.account.decimals
```
fluctlight-kayaba, line 17: promote `10 ** fromToken.account.decimals` into a variable, to improve readability.

apps/wallet/src/features/Swap/InputSwap/FromToken.tsx
```diff
@@ -0,0 +1,131 @@
...
+	};
+
+	const updateAmount = (amount: string) => {
+		swapActions.update({ amount });
+	};
+
+	const handlePressMax = () => {
+		if (!fromToken) return;
+		const maxAmount =
+			parseFloat(fromToken.account.balance) / 10 ** fromToken.account.decimals;
```
fluctlight-kayaba, line 39: group `10 ** fromToken.account.decimals;` under variable or `()` to be more declarative.

apps/wallet/src/features/Swap/InputSwap/ToToken.tsx
```diff
@@ -0,0 +1,102 @@
...
+	const handleSelectToToken = () => {
+		swapActions.openSelectToken('to');
+	};
+
+	const outAmount = swapQuote
+		? (parseInt(swapQuote.outAmount) * 1.0) / 10 ** (toToken?.decimals || 0)
+		: 0;
+
+	const updateSwapQuote = useCallback(
+		throttle(
```
fluctlight-kayaba, line 29: use `use-debounce` library for this, traditional `useCallback` + `throttle` did not work well in React Native, see [this line/file](https://github.cl/cocrafts/walless/blob/5dc1b4daa00788e0f90dc36ecac1c186fa58aa33/apps/mobile/src/utils/hooks/navigation.ts#L5) as reference.

apps/wallet/src/features/Swap/Select/SelectFromToken.tsx
```diff
@@ -0,0 +1,80 @@
...
+	const { tokens } = useTokens(network);
+
+	const filteredTokens = useMemo(() => {
+		const search = searchText.toLowerCase();
+
+		return tokens.filter((t) => {
+			return (
+				t.account.mint.toLowerCase().includes(search) ||
+				(t.metadata?.name || '').toLowerCase().includes(search) ||
+				(t.metadata?.symbol || '').toLowerCase().includes(search)
```
fluctlight-kayaba, line 25: promote each branch/case as a variable with self-explained names if possible.

apps/wallet/src/features/Swap/Select/SelectFromToken.tsx
```diff
@@ -0,0 +1,80 @@
...
+			);
+		});
+	}, [searchText]);
+
+	const handleBack = () => {
+		swapActions.closeSelectToken('from');
+	};
+
+	const handleSelectToken = (token: TokenDocument) => {
+		if (token._id != fromToken?._id) {
```
fluctlight-kayaba, line 35: use `!==`

apps/wallet/src/features/Swap/Success.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { swapContext } from './context';
+
+const Success = () => {
+	const insets = useSafeAreaInsets();
+	const containerStyle: ViewStyle = {
+		...styles.container,
+		paddingTop: Math.max(insets.top, 20),
+		paddingBottom: 20,
+	};
```
fluctlight-kayaba, line 15: this is in-efficient, use `containerStyle = [styles.container, { ...dyanmicProps }];` instead.

apps/wallet/src/features/Swap/Success.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+	const { fromToken, amount, toToken, swapQuote } =
+		useSnapshot(swapContext).swap;
+
+	const fromSymbol = fromToken?.metadata?.symbol;
+	const toSymbol = toToken?.symbol;
+	const outAmount = swapQuote
+		? (parseInt(swapQuote.outAmount) * 1.0) / 10 ** (toToken?.decimals || 0)
+		: 0;
```
fluctlight-kayaba, line 24: this kind of logic start to repeats, consider an util method for this kind of logic.

apps/wallet/src/utils/hooks/jupiter.ts
```diff
@@ -0,0 +1,37 @@
...
+const jupiterContext = proxy<JupiterContext>({
+	tokens: [],
+});
+
+export const useJupiterContext = () => {
+	const ctx = useSnapshot(jupiterContext);
+
+	return useMemo(() => {
+		if (ctx.tokens.length === 0) {
+			fetch('https://token.jup.ag/strict').then(async (res) => {
```
fluctlight-kayaba, line 30: promote as environment variable or as Firebase remote-config

apps/wallet/src/utils/transaction/swap/jupiter.ts
```diff
@@ -0,0 +1,95 @@
+import { VersionedTransaction } from '@solana/web3.js';
+import { logger } from '@walless/core';
+import type { TokenDocument } from '@walless/store';
+
+import type { SwapQuote } from './types';
+
+export const getMappedMint = (token: TokenDocument) => {
```
fluctlight-kayaba, line 7: will `getAliasedMint` may sound more relevant?

apps/wallet/src/utils/transaction/swap/jupiter.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export const getSwapQuote = async ({
+	fromMint,
+	toMint,
+	amount,
+	slippageBps = 50,
+}: GetSwapParams): Promise<SwapQuote | undefined> => {
+	const res = await fetch(
+		`https://quote-api.jup.ag/v6/quote?inputMint=${fromMint}&outputMint=${toMint}&amount=${amount}&slippageBps=${slippageBps}`,
```
fluctlight-kayaba, line 33: consider promote this as environment variable or Firebase remote-config.

apps/wallet/src/utils/transaction/swap/jupiter.ts
```diff
@@ -0,0 +1,95 @@
...
+		fromMint,
+		toMint,
+		amount,
+		slippageBps,
+	});
+	if (!quoteResponse) {
+		throw Error('Can not fetch swap quote, please try again!');
+	}
+
+	const res = await fetch('https://quote-api.jup.ag/v6/swap', {
```
fluctlight-kayaba, line 70: could be envrionment variable.

apps/wallet/src/utils/transaction/swap/jupiter.ts
```diff
@@ -0,0 +1,95 @@
...
+		body: JSON.stringify({
+			// quoteResponse from /quote api
+			quoteResponse,
+			// user public key to be used for the swap
+			userPublicKey,
+			// auto wrap and unwrap SOL. default is true
+			wrapAndUnwrapSol,
+			// feeAccount is optional. Use if you want to charge a fee.  feeBps must have been passed in /quote API.
+			// feeAccount: "fee_account_public_key"
+		}),
```
fluctlight-kayaba, line 84: a single comment with link to documentation page, or api docs would be better -- as the idea/meaning would be change externally

apps/wallet/src/features/Swap/InputSwap/FromToken.tsx
tanlethanh: `()` is automatically removed by code linter

apps/wallet/src/features/Swap/InputSwap/SelectButton.tsx
```diff
@@ -0,0 +1,60 @@
+import { type FC, Fragment } from 'react';
```
ltminhthu, line 1: `import type { FC } from 'react'`

apps/wallet/src/features/Swap/ConfirmPasscode/index.tsx
devtronghieu, line 1: We already have this file in `features/Send`, right? Can we reuse it?

apps/wallet/src/utils/transaction/swap/jupiter.ts
```diff
@@ -0,0 +1,91 @@
+import { VersionedTransaction } from '@solana/web3.js';
+import { logger } from '@walless/core';
+import type { TokenDocument } from '@walless/store';
+import { environment } from 'utils/config';
+
+import type { SwapQuote } from './types';
+
+export const getAliasedMint = (token: TokenDocument) => {
+	if (token.account.mint === '11111111111111111111111111111111') {
```
ltminhthu, line 9: can we use `solMint` here instead of `1111111111111...11111`

packages/gui/components/SwipeDownGesture.tsx
```diff
@@ -90,6 +90,7 @@ export default SwipeDownGesture;
 
 const styles = StyleSheet.create({
 	container: {
+		flex: 1,
 		flexGrow: 1,
```
ltminhthu, line 94: can we remove `flexGrow: 1` here?

apps/wallet/src/features/Swap/InputSwap/ToToken.tsx
```diff
@@ -0,0 +1,101 @@
...
+		paddingBottom: 30,
+	},
+	title: {
+		color: '#FFFFFF',
+	},
+	tokenContainer: {
+		flexDirection: 'row',
+		justifyContent: 'space-between',
+	},
+	outAmount: {
```
devtronghieu, line 97: may rename it to `outAmountText`

apps/wallet/src/features/Swap/InputSwap/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+			showError({ errorText: (error as Error).message });
+		}
+		setLoading(false);
+	};
+
+	const handleClose = () => {
+		swapActions.closeSwap();
+		setTimeout(() => {
+			swapActions.resetContext();
+		}, 200);
```
devtronghieu, line 37: why do we have to wait for 200ms to reset the context? it seems like a magic number

apps/wallet/src/features/Swap/context.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const initialContext = {
+	amount: '',
+};
+
+export const swapContext = proxy<{ swap: SwapContext }>({
+	swap: initialContext,
+});
+
+const modalMap = {
```
devtronghieu, line 32: may we make it more "typescript"?

```ts
enum SwapToken {
  From,
  To,
}

interface SwapModal {
  id: string;
  component: ReactNode;
}

const modalMap: Record<SwapToken. SwapModal> = {};
```

Because I see that you have use the `"from"` or `"to`" in other components, we can reuse this `enum`

apps/wallet/src/modals/Swap.tsx
```diff
@@ -0,0 +1,64 @@
...
+type Props = {
+	config: ModalConfigs;
+	props: SwapProps;
+};
+
+const SwapModal: FC<Props> = ({ config, props }) => {
+	const insets = useSafeAreaInsets();
+
+	const safeAreaStyle: ViewStyle = {
+		marginTop: insets.top || 20,
```
devtronghieu, line 27: I have seen in one of the above files, you applied the `Math.max()` to get the this number, may it use here?

apps/wallet/src/utils/transaction/swap/jupiter.ts
```diff
@@ -0,0 +1,91 @@
+import { VersionedTransaction } from '@solana/web3.js';
+import { logger } from '@walless/core';
+import type { TokenDocument } from '@walless/store';
+import { environment } from 'utils/config';
+
+import type { SwapQuote } from './types';
+
+export const getAliasedMint = (token: TokenDocument) => {
```
devtronghieu, line 8: Does this function convert a `devnet` mint address to a `mainnet-beta` mint address? I don't get the function name but inferring from its implementation

packages/icons/components/Switch.tsx
```diff
@@ -0,0 +1,17 @@
+import type { FC } from 'react';
+import { Path, Svg } from 'react-native-svg';
+
+import type { IconProps } from './types';
+
+export const Switch: FC<IconProps> = ({ size = 24, color = '#FFFFFF' }) => {
+	return (
+		<Svg width={size} height={size} viewBox="0 0 17 19" fill="none">
```
devtronghieu, line 8: just my preference, when the `viewBox` is not a square (in this case, `17` in width and `19` in height or vice versa), I will throw it inside a square frame (which may be 19x19) in Figma and then expose that frame again to get a square icon xD

apps/wallet/src/features/Swap/ConfirmPasscode/index.tsx
tanlethanh: The `Passcode` component is small enough for reusable purpose, It might be better to decouple the code instead of trying to reuse it strictly

apps/wallet/src/features/Swap/InputSwap/ToToken.tsx
tanlethanh: it is not necessary to rename, `name` and `type` for this purpose, we just need to add suffix like `text` in case it is confusing or has the same name with a different var

apps/wallet/src/features/Swap/InputSwap/index.tsx
tanlethanh: yep, It's a magic number. But the problem comes from the animation behavior of the modal component, it should expose life cycle methods to control when to remove context correctly. I keep the code, but will mark it to resolve in an issue

apps/wallet/src/features/Swap/context.ts
tanlethanh: It has been 'typescript' already, typescript allows us to use values as type, in case it needs, a little simple values, I prefer keeping the values as type

apps/wallet/src/modals/Swap.tsx
tanlethanh: These margin/padding values are used for different purpose

apps/wallet/src/utils/transaction/swap/jupiter.ts
tanlethanh: No, this function primarily maps so; to wrapped sol which is used to handle as sol in `solana program`. I also add a dev case to easily test the flow (because swap API endpoint does not have any dev tool'

packages/icons/components/Switch.tsx
tanlethanh: :v I just keep the icon viewBox, it is not necessary to wrap it into a square box, width, and height do their job, viewBox is used for the relative size of child lines

apps/wallet/src/utils/transaction/swap/jupiter.ts
devtronghieu: You may add some comments to brief this purpose in case someone else may work with it in the future

FME849 (approved):

PR #460, opened 2024-02-15
[WL-966] fix: `Widget not found` error after navigated to `Collection` or `Collectible` screens

FME849:
Ignoring all routes excluded `Widget` screen when updating new `latestLocation`

### DIFF
--- a/apps/wallet/src/utils/hooks/navigation.ts
+++ b/apps/wallet/src/utils/hooks/navigation.ts
@@ -26,12 +26,9 @@ export const useNavigationHydrate = () => {
 			const screen = widget?.screen as any;
 			const widgetId = widget?.params?.id;
 			/* eslint-enable */
+			const isWidgetScreen = screen === 'Widget';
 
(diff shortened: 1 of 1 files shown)


apps/wallet/src/utils/hooks/navigation.ts
```diff
@@ -23,10 +23,15 @@ export const useNavigationHydrate = () => {
...
 			const widget = childRoute?.params as any;
+			const screen = widget?.screen as any;
 			const widgetId = widget?.params?.id;
 			/* eslint-enable */
 
-			if (childRoute?.name === 'Explore' && widgetId !== lastWidgetId) {
+			if (
+				childRoute?.name === 'Explore' &&
+				screen === 'Widget' &&
+				widgetId !== lastWidgetId
```
tanlethanh, line 33: You could construct a new variable with a meaningful name to make it clear

tanlethanh (approved):

PR #467, opened 2024-02-26
[WL-915] refactor:  navigation system

FME849:


apps/wallet/src/components/StackContainer/index.tsx
```diff
@@ -13,7 +13,7 @@ import StackHeader from './Header';
 interface Props {
 	title: string;
 	style?: StyleProp<ViewStyle>;
-	noHeader?: boolean;
+	headerActive?: boolean;
```
ltminhthu, line 16: should be `isHeaderActive` or you can keep it the same (`noHeader`)

ltminhthu (approved):

apps/wallet/src/components/StackContainer/index.tsx
tanlethanh: I see people use `enableHeader` or `showHeader` @ltminhthu :v

apps/wallet/src/features/Home/index.tsx
```diff
@@ -45,8 +69,18 @@ const styles = StyleSheet.create({
 		alignItems: 'center',
 		gap: 36,
 	},
+	headerContainer: {
+		width: '100%',
```
tanlethanh, line 73: try to not use `width: '100%'`

apps/wallet/src/stacks/Explorer/ProfileStack/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const ProfileStack: FC<Props> = () => {
+	const screenOptions: StackNavigationOptions = {
+		headerShown: false,
+	};
+
+	const ManageProfileScreen = useMemo(
+		() =>
+			withStackContainer(ProfileScreen, {
+				title: '',
```
tanlethanh, line 27: remove this props if it's a empty string

tanlethanh (approved):

apps/wallet/src/components/StackContainer/index.tsx
sle-hm: vote for @ltminhthu suggestion

apps/wallet/src/features/Explorer/index.tsx
```diff
@@ -18,12 +18,14 @@ import ExplorerWidgetItem from './WidgetItem';
 interface Props {
 	style?: StyleProp<ViewStyle>;
 	widgets?: WidgetDocument[];
+	headerActive?: boolean;
```
sle-hm, line 21: proposal: `isHeaderActive` or `noHeader`

apps/wallet/src/features/Explorer/index.tsx
```diff
@@ -18,12 +18,14 @@ import ExplorerWidgetItem from './WidgetItem';
...
 	onToggleDrawer?: () => void;
 }
 
 export const ExplorerFeature: FC<Props> = ({
 	style,
 	widgets = mockWidgets,
+	headerActive = false,
```
sle-hm, line 28: `isHeaderActive` or `noHeader`

apps/wallet/src/features/Home/History.tsx
```diff
@@ -2,18 +2,33 @@ import { type FC, useEffect, useState } from 'react';
...
 
 	const handleNavigateToHistory = () => {
+		if (isMobileDisplay) {
+			navigate('Dashboard', {
+				screen: 'Home',
+				params: {
+					screen: 'History',
+				},
+			});
+			return;
```
sle-hm, line 22: avoid this kind of `return;` line.. keep it consistent by `if`.. `else`

apps/wallet/src/features/Home/index.tsx
```diff
@@ -2,27 +2,51 @@ import type { ViewStyle } from 'react-native';
...
+		navigate('Dashboard', {
+			screen: 'Explore',
+			params: {
+				screen: 'Profile',
+				params: {
+					screen: 'Setting',
+				},
+			},
+		});
+	};
```
sle-hm, line 36: wonder if we can make this method reuseable? seems that it could be a static constant (no hook or effect values being used, and the setting screen also well defined without component level context)

apps/wallet/src/features/Home/index.tsx
sle-hm: remove `alignItems: center` in the container `View`, then we don't need this `width: 100%`.
to align center something, in container level.. put that chunk of element under another wrapped `View` that itself do `alignItems: center`, or if there is only one single item needed to be center use `alignSelf: center` on it.

apps/wallet/src/features/Widget/BuiltInNetwork/CollectiblesTab.tsx
```diff
@@ -12,7 +12,7 @@ interface Props {
 
 export const CollectiblesTab: FC<Props> = ({ collections = [] }) => {
 	const { onGridContainerLayout, width } = useLazyGridLayout({
-		referenceWidth: 156,
+		referenceWidth: 150,
```
sle-hm, line 15: wonder why need this change?

apps/wallet/src/features/Widget/index.tsx
```diff
@@ -1,40 +1,36 @@
...
+import { useMemo } from 'react';
+import { withStackContainer } from 'components/StackContainer';
 
 import { extractWidgetComponent } from './internal';
 
 interface WidgetFeatureProps {
 	widgetId: string;
 	title: string;
-	style?: ViewStyle;
+	headerActive?: boolean;
```
sle-hm, line 10: `isHeaderActive` or `noHeader`

apps/wallet/src/features/Widget/index.tsx
```diff
@@ -1,40 +1,36 @@
...
 	onToggleDrawer: () => void;
 }
 
 export const WidgetFeature: FC<WidgetFeatureProps> = ({
 	widgetId,
 	title,
-	style,
+	headerActive = false,
```
sle-hm, line 17: `isHeaderActive` or `noHeader`, and sounds like we got duplicated component/block?

apps/wallet/src/index.tsx
```diff
@@ -14,24 +15,33 @@ import { linking, navigationRef } from 'utils/navigation';
...
-					theme={theme}
-					linking={linking}
-					onReady={hydrate.onNavigationReady}
-					onStateChange={hydrate.onNavigationStateChange}
+			<View style={[styles.container, !isMobileDisplay && styles.centerLayout]}>
+				<View
+					style={[
+						styles.innerLayout,
+						!isMobileDisplay && styles.extensionLayout,
+					]}
```
sle-hm, line 31: for long block like this, if possible promote it as a variable outside XML/Markup block, e.g:
```typescript
const innerStyle: StyleProp<ViewStyle> = [
  styles.innerLayout,
  !isMobileDisplay && styles.extensionLayout,
];

<View ...>
  <View style={innerStyle}>
</View>
```

apps/wallet/src/stacks/Explorer/WidgetNavigator/index.tsx
```diff
@@ -103,12 +106,12 @@ export const WidgetNavigator: FC<Props> = ({
 				})}
 			</View>
 
-			{runtime.isBrowser && profile?.profileImage && (
+			{navigationDisplay.sidebarAvatarActive && profile?.profileImage && (
```
sle-hm, line 109: should be `isSidebarAvatarActive`

apps/wallet/src/stacks/Explorer/index.tsx
```diff
@@ -1,23 +1,26 @@
...
 export const ExplorerStack = () => {
+	const { navigationDisplay } = useSnapshot(appState);
 	const screenOptions: DrawerNavigationOptions = {
 		headerShown: false,
 		drawerStyle: styles.drawer,
 		swipeEdgeWidth: 5000,
 		swipeMinDistance: sidebarWidth / 3,
 		overlayColor: 'transparent',
-		drawerType: runtime.isExtension ? 'permanent' : 'back',
+		drawerType: navigationDisplay.drawerPermanent ? 'permanent' : 'back',
```
sle-hm, line 23: should be `isPermanentDrawer` or `forcePermanentDrawer` or `usePermanentDrawer`

apps/wallet/src/state/app.ts
```diff
@@ -35,10 +44,44 @@ export const appState = proxy<AppState>({
...
+	const isMobileLayout =
+		!runtime.isExtension && (isResponsiveWidth || runtime.isMobile);
+	appState.isMobileDisplay = isMobileLayout;
+
+	if (isMobileLayout) {
+		appState.navigationDisplay = {
+			navigationHeaderActive: true,
+			drawerPermanent: false,
+			bottomTabActive: true,
+			sidebarAvatarActive: false,
```
sle-hm, line 75: for boolean/flag variable, highly recommend `is..` prefix - e.g `isPermanentDrawer`.

apps/wallet/src/state/bootstrap.ts
```diff
@@ -45,6 +45,7 @@ export const bootstrap = async (): Promise<void> => {
 	]);
 
 	configureDeviceAndNotification();
+	controlNavigationDisplay();
```
sle-hm, line 48: should not trigger this one here, just call it right in side `apps/wallet/src/state/app.ts` after declaration, or make sure inital values are accurate; put this method here cause unessary complexity/confusion.. I guess.

apps/wallet/src/features/Widget/BuiltInNetwork/CollectiblesTab.tsx
FME849: The `width` returned from `useLazyGridLayout` in extension layout is always `151px`. The calculated width need to be smaller than the reference to keep the layout in CollectiblesTab has 2 columns, therefore we just have to pick a reference value that smaller than calculated. The responsiveness in small view is the same.

apps/wallet/src/features/Home/index.tsx
FME849: this setting btn only shows up in extension thus this route is not available in mobile layout. Therefore I find it's not usable anywhere else.

PR #472, opened 2024-02-28
[WL-873] Balance error modal

devtronghieu:


apps/wallet/src/utils/transaction/common.ts
```diff
@@ -404,3 +405,48 @@ export const getNetworkMetadata = (network: Networks) => {
...
+): boolean => {
+	switch (network) {
+		case Networks.solana:
+			return tokenSymbol === 'SOL';
+		case Networks.sui:
+			return tokenSymbol === 'SUI';
+		case Networks.aptos:
+			return tokenSymbol === 'APT';
+		case Networks.tezos:
+			return tokenSymbol === 'XTZ';
```
fluctlight-kayaba, line 421: is it safe enough to check via `tokenSymbol`? I think better use `tokenAdress`.

apps/wallet/src/utils/transaction/common.ts
```diff
@@ -404,3 +405,48 @@ export const getNetworkMetadata = (network: Networks) => {
...
+		default:
+			return false;
+	}
+};
+
+export const hasEnoughBalanceToMakeTx = async (payload: TransactionPayload) => {
```
fluctlight-kayaba, line 428: would be very nice to apply/write unit test for these kind of methods.

apps/wallet/src/features/Send/ConfirmByPasscode/index.tsx
```diff
@@ -52,22 +56,29 @@ const PasscodeInput: FC<Props> = ({ navigator, item, activatedId }) => {
...
-					setError('Wrong passcode');
-				} else if (res.responseCode == ResponseCode.SUCCESS) {
-					const signature =
-						res.signatureString || res.signedTransaction?.digest || res.hash;
-					txActions.update({ signatureString: signature });
-					navigator.slideNext();
+				const hasEnoughBalance = await hasEnoughBalanceToMakeTx(payload);
+
+				if (!hasEnoughBalance) {
+					showError({ errorText: 'Insufficient balance to send' });
```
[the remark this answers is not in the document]
devtronghieu: it should be in `try` .. `catch` block as it called `getTransactionFee` and `getTransactionAbstractFee` inside its implementation, these two functions may throw error

apps/wallet/src/utils/transaction/common.ts
devtronghieu: It is much better using the `tokenAddress`, the above code is following the checking condition of `constructTransaction` in the same file. I don't think I will resolve this checking approach in this #PR, may I add a `// TODO` and create another ticket to resolve this issue?

tanlethanh: please use `tokenAddress`, it might cause breaking in the future

tanlethanh (approved):

PR #477, opened 2024-02-29
fix: handle rpc limit on solana in production environment

tanlethanh:


apps/wallet/browser/kernel/pwa.ts
```diff
@@ -3,7 +3,7 @@ import { logger } from '@walless/core';
 export const configurePWA = () => {
 	self.addEventListener('install', onInstall);
 	self.addEventListener('activate', onActivate);
-	self.addEventListener('fetch', onFetch);
+	// self.addEventListener('fetch', onFetch);
```
fluctlight-kayaba, line 6: mark as `TODO:` as definitely we'll need to comeback to this

apps/wallet/src/engine/runners/solana/collectibles.ts
```diff
@@ -87,22 +93,17 @@ export const loadCollectibleMetadata = async (
...
-	let nft = await mpl.nfts().load({
-		metadata: metadata as Metadata<JsonMetadata<string>>,
-		tokenOwner: owner,
-	});
-
-	if (!metadata.json && 'mintAddress' in metadata) {
-		const nftByMint = await mpl.nfts().findByMint({
-			mintAddress: metadata.mintAddress,
+	const nft = {
+		...(await mpl.nfts().load({
```
fluctlight-kayaba, line 97: I guess we don't need to clone this object, as no one else using/reference it

tanlethanh: actually, the obj return by findByMint is frozen and is read-only

apps/wallet/src/engine/runners/solana/history.ts
```diff
@@ -189,6 +203,7 @@ const getSplTransactionBalances = async (
 	);
 
 	const token: TransactionHistory['token'] = {
+		cluster: '',
```
fluctlight-kayaba, line 206: would it be okay to have empty `''` cluster?

fluctlight-kayaba (approved):

PR #493, opened 2024-03-13
feat: add migration script for storage v2

tanlethanh:


packages/store/migrations/index.ts
```diff
@@ -0,0 +1,102 @@
...
+) => {
+	logger.info(`start migrating database, scope: ${scope}`);
+	const setting = await storage.safeGet<SettingDocument>('settings');
+	if (!setting?.profile?.id) return;
+
+	const storedVersion = setting?.config?.storageVersion || 0;
+	const latestVersion = migrations[migrations.length - 1].version;
+
+	if (storedVersion < latestVersion) {
+		const newerFilter = (i: Migration) =>
```
FME849, line 26: I think `filterNewMigrations` or `filterVersionMigrations` have more transparent meaning

packages/store/migrations/index.ts
```diff
@@ -0,0 +1,102 @@
...
+						await storage.removeDoc(token._id);
+					});
+				});
+
+			const migrateNFTPromises = storage
+				.find<PouchDocument<unknown>>({
+					selector: { type: 'NFT', account: { $exists: true } },
+				})
+				.then(({ docs: tokens }) => {
+					return tokens.map(async (token) => {
```
FME849, line 69: do we need return value here?

packages/store/utils/type.ts
```diff
@@ -19,6 +19,11 @@ import type {
...
 } from '@walless/core';
 
+import type helpers from '../plugins';
+
+export type Database = Omit<PouchDB.Database, 'find'> &
+	typeof helpers & { find: TypedFind };
+
 export type DocumentType =
 	| 'Setting'
 	// | 'EndpointMap'
```
FME849, line 29: Do we still need this? If not, remove it

packages/store/migrations/index.ts
tanlethanh: yes, it need to return a list of promise to use `await Promise.all()`

FME849 (approved):

PR #500, opened 2024-03-19
feat: extension store blue argon remover via webpack

fluctlight-kayaba:
Google Extension store's review reject bundled `app.js` if there is tracking link inside the source code, no matter we use it or not; then, one of the suggested way is to remove/replace them with `''` to prove that our Extension compliance with a policy they called `bue argons`.


tanlethanh:
@fluctlight-kayaba is this included in v1.1.0 and have been submitted successfully

fluctlight-kayaba:
> @fluctlight-kayaba is this included in v1.1.0 and have been submitted successfully

yes, already included

PR #502, opened 2024-03-21
[WL-1094] feat: sui wallet standard

FME849:


apps/wallet/src/utils/storage/migrations.ts
```diff
@@ -0,0 +1,84 @@
...
+							},
+						);
+						await storage.upsert<PublicKeyDocument>(
+							suiPublicKeyDoc._id,
+							async (doc) => {
+								doc['meta'] = { encodedPublicKey };
+								return doc;
+							},
+						);
+						return {}; // Returning empty object just to satisfy type of onPasscodeComplete
```
tanlethanh, line 73: define new generic type for `PublicKeyDocument` to use for decoupling multiple networks

packages/store/migrations/index.ts
```diff
@@ -99,4 +134,15 @@ const migrations: Migration[] = [
...
+		version: 3,
+		description:
+			"This migration add encodedPublicKey to Sui PublicKeyDocument['meta']",
+		scope: 'app',
+		migrate: async () => {},
+	},
 ];
+
+export const getLatestMigrationVersion = () => {
+	return migrations[migrations.length - 1].version;
```
tanlethanh, line 147: get by max value

tanlethanh (approved):

packages/store/migrations/index.ts
FME849: we agreed that migrations will contain all the migrate version, so I think this method still worked in the way we expected

PR #545, opened 2024-05-24
fix: check registered account for sign-in again (first run) & clean up

devtronghieu:


apps/wallet/src/modals/LogoutModal.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { Button, modalActions } from '@walless/gui';
+import { appActions } from 'state/app';
 import { logout } from 'utils/auth';
 import { resetRoute } from 'utils/navigation';
 
 import { ModalId } from './types';
 
 const handleLogout = async () => {
 	await logout();
+	appActions.cleanupAfterLogOut();
```
tanlethanh, line 11: put this line into `logout`

apps/wallet/src/screens/Authentication/Login/internal.ts
```diff
@@ -35,7 +36,11 @@ export const signIn = async (invitationCode?: string) => {
 		} else if (status === ThresholdResult.Missing) {
 			navigate('Authentication', { screen: 'Recovery' });
 		} else if (status === ThresholdResult.Ready) {
-			navigate('Dashboard');
+			if (appState.profile.email) {
+				navigate('Dashboard');
+			} else {
+				navigate('Authentication', { screen: 'Recovery' });
+			}
```
tanlethanh, line 43: I don't think this case could happen, can you explain this one

apps/wallet/src/utils/auth/signIn.ts
```diff
@@ -81,6 +82,25 @@ const signInWithPasscode = async (
 		}
 	}
 
+	if (status === ThresholdResult.Ready) {
```
tanlethanh, line 85: better to use `else if`

apps/wallet/src/utils/auth/signIn.ts
```diff
@@ -81,6 +82,25 @@ const signInWithPasscode = async (
...
+				registerAccountWithoutKey: Account;
+			}>(mutations.registerAccountWithoutKey);
+
+			userAccount = account;
+		}
+
+		if (!userAccount.identifier) {
+			handleInitFail?.();
+			return;
+		}
```
tanlethanh, line 101: 1. prefer putting this block code into a new function `registerOldAccount`
2. if in this context, sign in flow, you can get the recovery key and register to api after they input the recovery key

apps/wallet/src/screens/Authentication/Login/internal.ts
devtronghieu: This case happens when a user logs in, then logs out, and logs in again, the `status` is `ready` (the `requiredShares <= 0` and the `totalShares = 3`, it seems like our `storage.clearAllDocs()` func not works properly I guess, we can dive into it in another task). So in that case, we'll navigate directly to the `Dashboard` without storing any data.

tanlethanh: storage.clearAllDocs() only clears our storage usage, not including w3a. I think you need to resolve this one, it's an unexpected behavior and critical to resolve. Otherwise, I remember each user account has w3a metadata, this case might only happen when the signed in user is the one has just signed out. So you need to find a way that can clean w3a metadata of the signed user.

This flow needs to be clear, does not resolve by stack up the logic

apps/wallet/src/utils/auth/signIn.ts
devtronghieu: You're right, as we have asked for the recovery seed & reconstruct the `tkey` so we can call `initAndRegisterWallet` directly, we don't need to use the `registerWithoutKey` anymore. Do you want to revert some commits related to the `registerWithoutKey` in `gotenks` also (I'll remove it from `walless` also)?

tanlethanh: let's do it :v

tanlethanh (approved):

PR #563, opened 2024-06-13
sync staging from dev: tezos assets, update point system, minor UI change, correct auth flow

tanlethanh:


apps/wallet/src/engine/runners/tezos/index.ts
```diff
@@ -1,16 +1,96 @@
+import type { TezosToolkit } from '@taquito/taquito';
+import {} from '@taquito/tzip16';
```
FME849, line 2: remove empty import

PR #564, opened 2024-06-13
fix: remote duplicated Setting stack, prevent redundant nested widget…

tanlethanh:
… params (broken drawer active item)


apps/wallet/src/stacks/Home/index.tsx
```diff
@@ -1,4 +1,4 @@
-import type { FC } from 'react';
+import { type FC, useMemo } from 'react';
```
ltminhthu, line 1: `import type { FC } from 'React'`

apps/wallet/src/stacks/Setting/index.tsx
```diff
@@ -1,4 +1,4 @@
-import type { FC } from 'react';
+import { type FC, useMemo } from 'react';
```
ltminhthu, line 1: same as above

ltminhthu (approved):

PR #578, opened 2024-06-25
[wallet] sync: minor UI change & update point system & apply type-check & fix CI/CD

tanlethanh:


apps/wallet/src/components/StackContainer/Header.tsx
fluctlight-kayaba, line 1: shouldn't be included here

apps/wallet/src/engine/runners/solana/history/index.ts
```diff
@@ -160,7 +160,9 @@ const constructTransactionHistoryDocument = async (
 		date,
 		status,
 		...balances,
-	};
+	} as HistoryDocument<
```
fluctlight-kayaba, line 163: use `satisfy HistoryDocument<SolanaTransferHistory | SolanaSwapHistory | SolanaUnknownHistory
	>;` on this kind of case

apps/wallet/src/features/History/HistoryItem/index.tsx
```diff
@@ -13,15 +13,15 @@ import SolanaTransferHistoryItem from './SolanaTransfer';
 import { SolanaUnknownHistoryItem } from './SolanaUnknown';
 
 interface Props {
-	transaction: HistoryDocument;
+	transaction: HistoryDocument<
+		SolanaTransferHistory | SolanaSwapHistory | SolanaUnknownHistory
```
fluctlight-kayaba, line 17: we may want generic `History type?

apps/wallet/src/features/History/index.tsx
```diff
@@ -46,20 +46,22 @@ const HistoryFeature: FC<Props> = ({ style, network, limit }) => {
...
+								/>
+							) : (
+								<View
+									key={transaction.signature}
+									style={styles.dateAndTransactionContainer}
+								>
+									<Text style={styles.date}>{transactionDate}</Text>
+									<HistoryItem transaction={transaction} />
+								</View>
+							);
```
fluctlight-kayaba, line 64: not caused by this commit, but I don't think a `.map` render want to return `null`/`undefined` on its loops; as React Reconciler have no idea on how to optimize the rendering cycle.

apps/wallet/src/features/Loyalty/History/index.tsx
```diff
@@ -27,9 +27,9 @@ import ModalHeader from '../components/ModalHeader';
 
 import Item from './Item';
 
-interface PartnerQuestProps {
+type PartnerQuestProps = {
 	safeAreaInsets: EdgeInsets;
-}
+};
```
fluctlight-kayaba, line 32: Any specific reason to use `type` instead of `interface`? I think we'll need to clarify distinguish between `interface` and `type` - currently `interface` used quite random by my mistake. What I think, we'll stop using `interface` except polymorphism related places.

fluctlight-kayaba (approved):

apps/wallet/src/components/StackContainer/Header.tsx
tanlethanh: this is a sync PR for other PRs (from `dev` to `staging`), the type-checking refactoring was merged to `dev` to test CI/CD. Sr for this confusing. You could find it here https://github.com/cocrafts/walless/pull/577

apps/wallet/src/engine/runners/solana/history/index.ts
tanlethanh: I have just searched for `sastifies` vs `as`. `sastifies` is really better for type checking (actually be able to check be ts compiler) than `as` (we are responsible for type converting)

apps/wallet/src/features/Loyalty/History/index.tsx
tanlethanh: type is better for typecheck feature with typescript, but they are still overlap in their features. `interface` is only used for OOP inheritant, so I think we should use `type` as rich features of type checking, and `interface` for the case we want to apply OOP (just a convention).

apps/wallet/src/engine/runners/solana/history/index.ts
tanlethanh: I resolved by removing `as`, and also not use `satifies`. I think `satifies` stand for type definition of an object, and it was mentioned as return type of function which it is returned. Result actual type is much better for type safety instead of type assertion

PR #584, opened 2024-07-07
[wallet] fix: carousel force end gesture when mouse up, prevent stuck when moving out of gesture area

tanlethanh:


apps/wallet/src/features/Explorer/Highlights/CardCarousel.tsx
```diff
@@ -1,5 +1,6 @@
 import { type FC, useEffect, useRef } from 'react';
```
ltminhthu, line 1: change to `type { FC }`

ltminhthu, line 1: change `type { FC }`

tanlethanh: not in my change, but I will update it

ltminhthu (approved):

PR #594, opened 2024-07-09
[wallet][MS-245] first time user popup for adding pixeverse

ltminhthu:


apps/wallet/ios/Podfile.lock
```diff
@@ -993,4 +999,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 877abdf091993fa0cd980670b68c31cf31b2d28e
 
-COCOAPODS: 1.14.3
```
devtronghieu, line 996: I have no idea which version of pods that we should use, please double check this one

apps/wallet/src/modals/FirstTimePopup/PixeverseCard.tsx
```diff
@@ -0,0 +1,89 @@
+import { Image, StyleSheet, Text, View } from 'react-native';
```
devtronghieu, line 1: we will import the `Text` from `@walless/gui` instead of `react-native`

apps/wallet/src/modals/FirstTimePopup/index.tsx
```diff
@@ -0,0 +1,116 @@
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
+import { modalActions, Text } from '@walless/gui';
+import { BlueCircleBackground } from '@walless/icons';
```
devtronghieu, line 3: I don't think this is an icon, right?

Btw, you can create a `CircleBackground` and passing the `Blue` color as a prop

apps/wallet/src/modals/FirstTimePopup/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const FirstTimePopup = () => {
+	const handleAddPixeverse = () => {
+		addWidgetToStorage(mockWidgets[0]._id, mockWidgets[0]);
+		navigate('Dashboard', {
+			screen: 'Explore',
+			params: {
+				screen: 'Widget',
+				params: {
+					id: mockWidgets[0]._id,
```
devtronghieu, line 19: I don't think you should use the `[0]` index to query for `Pixeverse` widget, should use a concrete pixeverse id instead

apps/wallet/src/modals/FirstTimePopup/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+					<Text style={styles.buttonText}>Add Pixeverse</Text>
+				</TouchableOpacity>
+			</View>
+		</View>
+	);
+};
+
+const styles = StyleSheet.create({
+	container: {
+		width: 374,
```
devtronghieu, line 61: you should use `padding` or children's `margin` instead of fixed-size container

apps/wallet/src/state/app.ts
```diff
@@ -27,6 +27,7 @@ export interface AppState {
 		isBottomTabActive: boolean;
 		isSidebarAvatarActive: boolean;
 	};
+	showPopup: boolean;
```
devtronghieu, line 30: should be `showFirstTimePopup`

apps/wallet/src/state/bootstrap.ts
```diff
@@ -45,6 +46,8 @@ import { widgetState } from './widget';
 export const bootstrap = async (): Promise<void> => {
 	const startTime = new Date();
 	appState.remoteConfig = loadRemoteConfig();
+	appState.showPopup =
+		(await universalLocalStorage.getItem('showPopup')) ?? true;
```
devtronghieu, line 50: should create a var to hold this `showPopup` key

apps/wallet/src/modals/FirstTimePopup/PixeverseCard.tsx
```diff
@@ -0,0 +1,91 @@
+import { Image, StyleSheet, View } from 'react-native';
+import { Text } from '@walless/gui';
+import { mockWidgets } from 'state/widget';
+
+const PixeverseCard = () => {
+	const pixeverseWidget = mockWidgets.find((item) => item._id === 'pixeverse');
+	if (!pixeverseWidget) return null;
```
tanlethanh, line 7: this condition should be check outside this component, assume that this one is only work with available pixeverse widget data via prop

apps/wallet/src/modals/FirstTimePopup/PixeverseCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+		</View>
+	);
+};
+
+export default PixeverseCard;
+
+const styles = StyleSheet.create({
+	container: {
+		flexDirection: 'row',
+		width: 260,
```
tanlethanh, line 46: consider to use another approach instead of using hard code sizing

apps/wallet/src/modals/FirstTimePopup/PixeverseCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+	infoContainer: {
+		flex: 1,
+		justifyContent: 'space-between',
+	},
+	title: {
+		color: '#ffffff',
+		fontWeight: '500',
+	},
+	description: {
+		fontSize: 10,
```
tanlethanh, line 75: please check if fontSize 10 is valid for now, as I remember we refactored and changed whole font size to keep the app consistent (might be larger)

apps/wallet/src/utils/localStorage/index.ts
```diff
@@ -0,0 +1,27 @@
...
+	setItem: async (key, value) => {
+		if (typeof value === 'string') {
+			await AsyncStorage.setItem(key, value);
+		} else {
+			await AsyncStorage.setItem(key, JSON.stringify(value));
+		}
+	},
+	removeItem: async (key) => await AsyncStorage.removeItem(key),
+	clear: async () => await AsyncStorage.clear(),
+};
```
tanlethanh, line 27: we are using pouchDB engine as an universal storage for both mobile and web, please check to keep consistent, no fragmented

apps/wallet/src/utils/localStorage/index.web.ts
tanlethanh, line 1: same as above comment

apps/wallet/ios/Podfile.lock
```diff
@@ -841,6 +844,8 @@ EXTERNAL SOURCES:
     :path: "../../../node_modules/react-native/ReactCommon/runtimeexecutor"
   ReactCommon:
     :path: "../../../node_modules/react-native/ReactCommon"
+  RNCAsyncStorage:
+    :path: "../../../node_modules/@react-native-async-storage/async-storage"
```
tanlethanh, line 848: after reviewing you storage function, remove these changes if it is not needed anymore

apps/wallet/package.json
```diff
@@ -33,6 +33,7 @@
     "@metacraft/cli-web3-polyfills": "^0.0.7",
     "@metaplex-foundation/js": "0.20.1",
     "@mysten/sui.js": "0.51.1",
+    "@react-native-async-storage/async-storage": "^1.23.1",
```
tanlethanh, line 36: same as above comment

apps/wallet/src/utils/localStorage/index.ts
ltminhthu: @tanlethanh  but after logging out, the pouchDB would be cleared so I don't think we should use it to store the flag

tanlethanh: you need to handle to not clear the thing you want persist after logout

apps/wallet/src/modals/FirstTimePopup/PixeverseCard.tsx
ltminhthu: > please check if fontSize 10 is valid for now, as I remember we refactored and changed whole font size to keep the app consistent (might be larger)

i've just check, the fontSize 10 still works well

apps/wallet/ios/Podfile.lock
tanlethanh, line 1: still here, not resolve yet

apps/wallet/src/stacks/Dashboard/index.tsx
```diff
@@ -15,6 +25,24 @@ const Tab = createBottomTabNavigator<DashboardParamList>();
 
 export const DashboardStack = () => {
 	useNotificationPermissionRequest();
+	const { showFirstTimePopup: showPopup } = useSnapshot(appState);
```
tanlethanh, line 28: remove this alias

packages/store/plugins/clear.ts
```diff
@@ -1,8 +1,13 @@
 const persistedIds = ['system'];
 
-export async function clearAllDocs(this: PouchDB.Database) {
+export async function clearAllDocs(
+	this: PouchDB.Database,
+	whitelist: string[] = [],
```
tanlethanh, line 5: nonResetWhitelist

tanlethanh (approved):

tanlethanh (approved):

